PR #5, opened 2022-09-21
Full rewrite

### LINKED ISSUE #4: Numpy utilities
Currently this package is focused on operations over only a few elements (vectors, matrices).

Some operations are more efficient with numpy, e.g. transforming an entire array of vertices (instead of a single vertex) with a matrix. We should provide a module for that kind of code and make the numpy dependency optional using extras_require.

We can add the functions as the need comes up when working on https://github.com/pygfx/pygfx examples.

Korijn:
Completely revise this repo and library (it's unused and on 0.2.0 anyway)

* Closes #4
* Based on @almarklein's notes here: https://gist.github.com/almarklein/452951a636f7fe67ca26a95da567cbb6
* Adds CI

I recommend filtering out deleted files when reviewing this PR since I deleted pretty much everything:

image

# pygfx checklist

- [ ] Matrix4
  - [x] get_inverse(matrix)
  - [x] set(matrix)
  - [x] make_orthographic(left, right, top, bottom, near, far)
    - [ ] support depth ranges (-1..1 and 0..1)
  - [x] premultiply(matrix)
  - [ ] look_at(eye, target, up)
  - [x] multiply_matrices(matrix, matrix)
  - [x] compose(position, rotation, scaling)
  - [x] decompose(position, rotation, scaling)
  - [x] copy()
- [ ] Vector3
  - [x] clone()
  - [ ] add(vector)
  - [ ] sub(vector)
  - [ ] copy(vector)
  - [ ] multiply(vector)
  - [ ] add_scaled_vector(vector)
  - [ ] normalize()
  - [ ] negate()
  - [ ] project(camera)
  - [ ] unproject(camera)
  - [ ] multiply_scalar(scalar)
  - [ ] apply_quaternion(quaternion)
  - [ ] set_from_spherical()
  - [ ] distance_to(vector)
  - [ ] length()
  - [ ] dot(vector)
  - [ ] cross(vector)
  - [ ] set_from_matrix_position(matrix)
  - [ ] apply_matrix(matrix)
- [ ] Spherical
  - [ ] set_from_vector(vector)
  - [ ] make_safe()
- [x] Quaternion
  - [x] set_from_unit_vectors(vector, vector)
  - [x] clone()
  - [x] inverse()
  - [x] set_from_rotation_matrix()
  - [x] set_from_axis_angle(vector, angle)
  - [x] premultiply(quaternion)
  - [x] multiply(quaternion)
- [ ] Misc
  - [ ] aabb_to_sphere
  - [x] transform
  - [ ] transform_aabb


pylinalg/func/vector.py
```diff
@@ -0,0 +1,57 @@
...
+    if out is None:
+        out = np.empty_like(v1 if v1.size > v2.size else v2)
+    out[:] = v1 + v2
+    return out
+
+
+def vector_sub_vector(v1, v2, out=None):
+    if out is None:
+        out = np.empty_like(v1 if v1.size > v2.size else v2)
+    out[:] = v1 - v2
```
Korijn, line 14: can I call np.sub(v1, v2, out=out) ?

pylinalg/obj/base.py
```diff
@@ -0,0 +1,35 @@
...
+        return self._val.__iter__()
+
+    @property
+    def __array_interface__(self):
+        return self._val.__array_interface__
+
+    def __eq__(self, other):
+        return isinstance(other, self.__class__) and np.array_equal(
+            self._val, other._val
+        )
```
Korijn, line 35: would be nice if you could write pla.Point(1, 2, 3) == [1, 2, 3] ? Or np.array([1, 2, 3]) ?

almarklein:
This is awesome. Can't wait to see this in pygfx :D

pylinalg/func/vector.py
Korijn: resolved this by just making vector_sub_vector an alias for np.subtract since it has exactly the same signature

pylinalg/obj/base.py
Korijn: done, see `test_linalgbase_eq` in `tests/test_obj_point.py` for a demonstration of all the supported cases

pylinalg/func/matrix.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+matrix_inverse = np.linalg.inv
+
+
+def matrix_compose(translation, rotation, scaling, out=None):
+    from .quaternion import quaternion_to_matrix
+
+    if out is None:
+        out = np.identity(4, dtype=translation.dtype)
```
Korijn, line 121: better to use empty here

pylinalg/func/matrix.py
```diff
@@ -0,0 +1,147 @@
...
+            quaternion_to_matrix(rotation),
+            matrix_make_scaling(scaling),
+        ]
+    )
+    return out
+
+
+def matrix_decompose(matrix, translation=None, rotation=None, scaling=None):
+    if translation is None:
+        translation = np.zeros((3,), dtype=matrix.dtype)
```
Korijn, line 134: empty is better

pylinalg/func/matrix.py
```diff
@@ -0,0 +1,147 @@
...
+    translation[:] = matrix[:-1, -1]
+
+    if scaling is None:
+        scaling = np.zeros((3,), dtype=matrix.dtype)
```
Korijn, line 138: empty is better

pylinalg/func/matrix.py
```diff
@@ -0,0 +1,147 @@
...
+    scaling[:] = np.linalg.norm(matrix[:-1, :-1], axis=0)
+    if np.linalg.det(matrix) < 0:
+        scaling[0] *= -1
+
+    if rotation is None:
+        rotation = np.array([0, 0, 0, 1], dtype=matrix.dtype)
```
Korijn, line 144: empty is better

pylinalg/func/vector.py
```diff
@@ -0,0 +1,65 @@
...
+        if the translation component should not be applied, 1 (positions)
+        otherwise.
+
+    Returns
+    -------
+    ndarray, [..., ndim]
+        transformed vectors
+    """
+    vectors = vector_make_homogeneous(vectors, w=w)
+    # the transpose is necessary due to the shape of the vectors matrix
```
Korijn, line 64: this could be clearer; because each vector is a _row_, not a _column_

Korijn:
This is quite a chore, so I'm going to try and limit the scope by doing the bare minimum work required to support pygfx in its current state. That's still a ton of work.

pylinalg/func/matrix.py
```diff
@@ -0,0 +1,154 @@
...
+        raise NotImplementedError()
+
+    matrix = np.identity(4, dtype=dtype)
+    matrix[0, 0] = factors[0]
+    matrix[1, 1] = factors[1]
+    matrix[2, 2] = factors[2]
+    return matrix
+
+
+def matrix_make_rotation_from_euler_angles(angles, order="xyz", dtype="f8"):
```
almarklein, line 49: Also need `matrix_make_rotation_from_axis_angle`?

pylinalg/obj/matrix.py
```diff
@@ -0,0 +1,71 @@
...
+    matrix_decompose,
+    matrix_inverse,
+)
+
+
+class Matrix(LinalgBase):
+    """A representation of a transformation matrix in 4D
+    homogeneous (to 3D Euclidean) space."""
+
+    def __init__(self, matrix=None, /, *, dtype=None):
```
almarklein, line 15: I know that the star makes that all args that come after it are keyword-only. Do I understand correctly that the slash there makes that the args come before it position-only?

pylinalg/obj/matrix.py
```diff
@@ -0,0 +1,71 @@
...
+                self._val = np.asarray(matrix, dtype=dtype)
+        else:
+            if dtype is None:
+                dtype = "f8"
+            self._val = np.identity(4, dtype=dtype)
+
+    def copy(self):
+        return self.__class__(self._val.copy())
+
+    def icompose(self, translation, rotation, scaling):
```
almarklein, line 29: "i" for internal, right?  On that note, I guess all these methods need a docstring :)

almarklein (commented):
I love how clean this code is ❤️

pylinalg/func/matrix.py
Korijn: Only the quaternion variant is in my checklist. I based it on a search of the pygfx repo.

pylinalg/obj/matrix.py
Korijn: Yes, that's correct

pylinalg/obj/vector.py
```diff
@@ -0,0 +1,72 @@
...
+        if isinstance(other, (int, float)):
+            new = Vector(0, 0, 0)
+            vector_mul_scalar(self._val, other, out=new._val)
+            return new
+        else:
+            if isinstance(other, LinalgBase):
+                # if not isinstance(other, Scalor):
+                #     raise TypeError(
+                #         "Can only multiply a Vector with a scalar or Scalor."
+                #     )
```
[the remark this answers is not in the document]
Korijn: For now I'm only implementing the things in the checklist, and I'll remove all the unrelated code. I like the idea you put forward but it's just too much effort in one go.

pylinalg/func/matrix.py
almarklein: 👍 can always be added later if someone needs it.

pylinalg/obj/matrix.py
almarklein: > Inplace

Ah, yes that's what I meant 😊

pylinalg/obj/vector.py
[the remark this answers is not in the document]
almarklein: 👍

Korijn:
> I love how clean this code is ❤️

After I've completed the checklist I'll look at all the functions, and see which conventions are feasible, apply them and document them.

Korijn:
New to do:

- [ ] Make CI pass
- [ ] Enforce conventions and document them
- [ ] Merge
- [ ] Create new issues for the remaining work

Korijn:
I've focused on deciding on conventions and ensuring CI passes. I also revamped `LinalgBase` to automatically wrap most of `ndarray` so we can lose a ton of boilerplate.

* @FirefoxMetzger and @almarklein can you review the `CONVENTIONS.md` document?

I intend to ensure the code that's in this branch complies with the conventions and sets the example, and then merge and create follow-up items.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+## Function signatures
+
+We strive to align closely with the conventions of [numpy](https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs)
+in order to be least-surprising for pylinalg users.
+
+All functions shall adhere as much possible to the following signature:
+
+`func(x1, x2, ..., /, *, out=None, dtype=None, **kwargs)`
+
+* All input arguments are positional-only.
```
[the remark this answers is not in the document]
FirefoxMetzger: That said, I would vote for extending this to 

```suggestion
`func(x1, x2, ..., /, *, out=None, dtype=None, **kwargs)`

* All input arguments are positional-only or keyword-only.
```

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+* There can be as many input arguments as needed for the function signature to make sense.
+* Input arguments are always [`array_like`](https://numpy.org/doc/stable/glossary.html?highlight=array_like#term-array_like):
```
FirefoxMetzger, line 55: I like the general idea behind this, but I fear it might be too restrictive. I can see scenarios where you want something like a `method="linear"` parameter or where you need a `eps=1e-6`. 

Of course, you can require that those are arrays and apply broadcasting to them to match the other inputs, but I am not sure if this adds value in comparison to making faulty inputs harder to catch.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+    * "Any scalar or sequence that can be interpreted as an ndarray. In
+      addition to ndarrays and scalars this category includes lists
+      (possibly nested and with different element types) and tuples.
+      Any argument accepted by `np.array` is `array_like`."
+    * TBD: require a call to `np.asarray` or `np.asanyarray` on inputs?
```
FirefoxMetzger, line 60: I am against `np.asanyarray`; it allows object arrays and you usually don't want these in a numerics library. `np.asarray` or `np.ascontiguousarray` make sense in different places, but I wouldn't require it on each function.

For future reference, I like this SO answer as an overview: https://stackoverflow.com/a/52103839/6753182

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+* All non-input arguments are keyword-only.
```
FirefoxMetzger, line 61: Ah, I see.  In light of this, we can either ignore my previous comment on keyword-only arguments or revise this statement. I think I still prefer my approach because it's a bit unclear to me what a non-input argument is.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+* If `out` is `None` (the default), a new return array is created. Otherwise,
+  the result of the function is written to `out`.
+    * If there are multiple outputs, `out` is expected to be a tuple
+      with matching number of elements.
```
FirefoxMetzger, line 65: This is not related directly to the guide, but `out` on transform functions can create problems, e.g. if I do thinks like 

```
gaussian_blur(image, 0.2, (5, 5), out=image)
```

Are we worrying about such scenarios here, or will we cross that bridge when we get there?

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+  dtypes chosen may depend on the function and the inputs may be cast
+  to this dtype to perform the calculation.
+* There may be additional keyword-only arguments as required by
+  the specifics of individual functions.
+
+Since the conventions align with those of numpy, in some cases, it is
+possible to simply alias a numpy function to avoid all overhead and
+implement the conventions. Optionally `functools.partial`
+can be used to limit the available arguments, or a small wrapper function
+can be created.
```
FirefoxMetzger, line 81: @Korijn Do you have an example of when this would be preferable over just calling the corresponding numpy function?

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+# Object oriented API conventions
+
+This API is for external use and for novice-users that want to
+perform linear algebra operations on conceptually familiar primitives.
+
+In this API each "thing" is represented as one object. 
+This API should make any linalg work much easier and safer, partly
+because semantics matters here: a point is not the same as a vector.
+
+These objects are array-like and iterable to make them easy to
```
FirefoxMetzger, line 93: Does this mean that one `Point` object can hold multiple points, or that that the coordinates (x, y, z) of a point are iterable and indexable?

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+
+The source for this API resides in the `pylinalg.obj` subpackage and is organized
+by type.
+
+## Imports cycles
+
+Since the classes here will regularly need to instantiate other types,
+circular import dependencies exist. To work around this, only the `LinalgBase`
+type can be imported at module level, and other types will have to be imported
+at runtime in methods.
```
FirefoxMetzger, line 108: This sounds a bit like our folder structure is suboptimal for what we aim to do. Is this an actual problem right now, or just a concern for the future?

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Function naming
+
+* Names should be concise and short.
+* For every method, there is an alternative in-place method, signified
+  with the prefix `i`. This is not the prettiest option, but it is
+  concise and short.
```
FirefoxMetzger, line 115: This seems contradictory to the aim of "Extensive input validation and ease of use is prioritized over performance.". 

There is no upside to in-place computation other than performance and if you get it wrong you end up with really hard-to-debug problems which is not very beginner friendly. I would vote for excluding in-place computation and reserving it for the functional/advanced API (which can do it via `method(x1, out=x1)`).

FirefoxMetzger (commented):
I left my thoughts on the conventions. I think most of them are pretty good. 

I did wonder though if we are really writing `ufunc`s here or if we are writing more general skimage-style functions that compute useful things on arrays. Depending on that we may wish to rephrase this text a bit.

CONVENTIONS.md
almarklein: I can imagine e.g. `Point.distance_to_plane()`  and `Plane.distance_to_point`, just to name one.

CONVENTIONS.md
almarklein: IIUC Args like these would be keyword arguments. A bullet further down says "There may be additionaleyword-only arguments as required by the specifics of individual functions."

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
+# Style conventions
+
+## Docstrings
+
+Docstrings shall be written in NumpyDoc format.
+
+Docstrings can either be just the summary text, or they must be complete.
```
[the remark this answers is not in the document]
Korijn: I agree with @FirefoxMetzger's suggestion here.

Just a heads up; I'm not going to actually write any docstrings in this PR, I feel like the effort would likely be wasted due to changes later. I want to wait a bit until we have a little more stability.

I will still put the convention in place.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Type annotations
+
+Type annotations will not be applied (for now).
+
+Note: I'm not very familiar with the latest and greatest type annotation
+options for libraries that make heavy use of numpy arrays like this one.
```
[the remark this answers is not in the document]
Korijn: Alright, that makes sense!

CONVENTIONS.md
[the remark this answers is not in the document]
Korijn: Yes, my bad. I will provide clarification and an example would probably also help.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+
+Performance is prioritized over extensive input validation.
+
+The source for this API resides in the `pylinalg.func` subpackage and is organized
+by type.
+
+## Function naming
+
+The functional API has rather verbose names, but it makes things
+explicit.
```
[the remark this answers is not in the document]
Korijn: I intended to imply all the general code style rules by stating which linter plugins and setup we use. It includes `pep8-naming`.

I'll make sure to call it out explicitly.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Function signatures
+
+We strive to align closely with the conventions of [numpy](https://numpy.org/doc/stable/reference/ufuncs.html#ufuncs-kwargs)
+in order to be least-surprising for pylinalg users.
```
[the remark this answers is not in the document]
Korijn: Sort of. I'll incorporate this change slightly differently.

CONVENTIONS.md
Korijn: Sure. `matrix_new = partial(np.identity, 4)` creates a function that creates identity affine transform matrices, while the other arguments of the underlying numpy function are still available (e.g. `out` and `dtype`).

You could also write:

```py
def matrix_new(*args, **kwargs):
    return np.identity(4, *args, **kwargs)
``` 

But that is more code, and also significantly slower. If you have this code in a hot code path, the `partial` will perform much better.

CONVENTIONS.md
Korijn: The latter.

CONVENTIONS.md
Korijn: Well, even simpler, `Point.to_vector()` and `Vector.to_point()` would be two functions that need to call each other's constructor. I think runtime imports are the proper way to deal with that.

CONVENTIONS.md
Korijn: There is another upside; element-specific mutation detection can be implemented efficiently only with in-place mutations.

I'm inclined to agree with you but I'm having a hard time seeing how pygfx's `WorldObject` attributes `position` etc would work when they would need to be replaced all the time.

CONVENTIONS.md
```diff
@@ -0,0 +1,125 @@
...
+    * If there are multiple outputs, `out` is expected to be a tuple
+      with matching number of elements.
+* If `dtype` is `None` (the default), the dtype of the return array
+  will be determined automatically by numpy based on the operations
+  performed on the inputs. Otherwise, the dtype of the output array(s) is
+  overridden with the
+  provided dtype with a call to `ndarray.astype`. This should ensure
+  a matching precision of the calculation. The exact calculation
+  dtypes chosen may depend on the function and the inputs may be cast
+  to this dtype to perform the calculation.
```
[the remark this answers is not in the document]
Korijn: I simplified this section. Essentially we just use numpy and don't bring anything else to the table. That means we default to numpy's default (float64) and you can override the data-types using the `out` and `dtype` arguments on your own if you want to.

CONVENTIONS.md
Korijn: Unit tests will have to ensure that those cases are covered properly, I think.

CONVENTIONS.md
Korijn: I revised it.

CONVENTIONS.md
Korijn: * I agree with `np.asarray` (and not `np.asanyarray`).
* Thanks for the SO reference. I know the differences, I'm just trying to reach consensus with you two on a decision. Thanks for the feedback, it helps make the right decision.

"`np.asarray` make sense in different places, but I wouldn't require it on each function." -> this is where the struggle lies; what is a good convention? When does it make sense, and when does it not make sense?

In my experience, it just gets added whenever people have the desire to call a function with a plain list instead of an ndarray, and the implementation just "accidentally" happens to perform an operation on that specific input that does not already implicitly call `np.asarray`.

CONVENTIONS.md
Korijn: Almar is correct. I tried to clarify this.

Korijn:
> I left my thoughts on the conventions. I think most of them are pretty good.

Thanks for the feedback, I will push my revised version shortly.

> I did wonder though if we are really writing `ufunc`s here or if we are writing more general skimage-style functions that compute useful things on arrays. Depending on that we may wish to rephrase this text a bit.

Definitely the latter. I only referred to the "ufunc" documentation since they provide a very elaborate explanation of the `out` and `dtype` keyword arguments, which we want to align with here.

CONVENTIONS.md
almarklein: FWIW `np.asarray` does not make a copy if the input already is an array. So it should not matter too much performance-wise.

CONVENTIONS.md
almarklein: @Korijn do you mean that the functions themselves will check for such cases and act appropriately? (i.e. not something the caller should worry about?)

CONVENTIONS.md
almarklein: Being able to assume that these objects are immutable also has favorable consequences (e.g. not having to worry about making a copy because the object might be modified without you knowing). But I agree that we need to see how this works out in PyGfx.

CONVENTIONS.md
Korijn: Yes

CONVENTIONS.md
FirefoxMetzger: > Thanks for the SO reference. I know the differences, I'm just trying to reach consensus with you two on a decision. Thanks for the feedback, it helps make the right decision.

Sorry, that may have come across differently from what I intended. I was thinking of this reference as a useful tool to discuss when we want to use which variant :)

> This is where the struggle lies; what is a good convention? When does it make sense, and when does it not make sense?

I agree with @almarklein here, that it is probably safe to add in most cases, since it is indeed a no-op if the array already is a base `ndarray` or a view and it is indeed an easy way to support ArrayLike input.

At the same time, there are two scenarios that come to mind where `np.asarray` is counter-productive. First, when it is important to preserve array subtypes, e.g. for sparse arrays, masked arrays, or if you have metadata (like ImageIOs old `image` class). Second, if you have an object whose buffer you could share with the base array, ie., where it is sufficient to create a view into a buffer instead of a dedicated array (I would have to check numpy's [`__array__` magic](https://numpy.org/devdocs/user/basics.dispatch.html) again to read up on if this can be solved elegantly).

Both cases can be handled separately though, so I would still opt for using `np.asarray` by default. That said, I would probably not make this a requirement in the conventions or at least state that there may be exceptions for special circumstances (like the above).

CONVENTIONS.md
FirefoxMetzger: What I was wondering is how `x = matrix_new()` improves `x=np.identity(4)`. Am I guessing correctly that it is meant for better readability?

CONVENTIONS.md
FirefoxMetzger: Got it. In that case, maybe we could say the following to be more precise?

```suggestion
These objects are backed by an array-like structure and are thus easy to
```

CONVENTIONS.md
Korijn: I guess to me it's similar to how it is can be easier for some users when everything they need is provided by `pygfx` and they don't need to also work with `wgpu`. :)

CONVENTIONS.md
FirefoxMetzger: Fair point. Could there be a middle ground by saying that objects are mutable in general, but that methods return copies? This would allow us to do in-place operations via the functional API (e.g. using `out=`) giving us access to performance while preserving the "validated and safe" aim of the object API.

I also like this because it makes it explicit where you could run into trouble: You are switching to the functional API here, so you need to pay extra attention to what you are doing.

FirefoxMetzger:
> I left my thoughts on the conventions. I think most of them are pretty good.

I just noticed my own spelling mistake 🤣 @Korijn I meant that most of the convention that you wrote looks pretty good.

CONVENTIONS.md
Korijn: Why didn't I think of that? Will apply.

CONVENTIONS.md
```diff
@@ -0,0 +1,172 @@
...
+            return np.dot(vectors, transform, out=out)
+        except ValueError:
+            # otherwise we need a temporary array and cast
+            out[:] = np.dot(vectors, transform)
+            return out
+    # otherwise just return whatever dot computes
+    out = np.dot(vectors, transform)
+    # cast if requested
+    if dtype is not None:
+        out = out.astype(dtype)
```
Korijn, line 120: `copy=False`

CONVENTIONS.md
Korijn: Illustrative example, I think:

```py
camera.position.z = 250
```

Would become:

```py
camera.position = camera.position + [0, 0, 250]
```

I don't think that's an improvement from the user perspective.

Korijn:
Alright, I completely redid most of the code, to clean it up significantly and to align with the conventions.

I propose to merge it as it is now and then re-asses the state and create tickets for any follow-up work.

@almarklein  @FirefoxMetzger  agreed?

almarklein:
Yeah, no problem for this code to be broken on master right now. And easier to move forward from there.

Korijn:
It's not broken! :)

PR #22, opened 2022-12-08
Add stubs for missing pygfx functions

FirefoxMetzger:
I've written stubs for all the functions that are currently missing to get pygfx compatibility. My aim is to get this in place so that we are clear on how each function should behave and to then implement each one in turn.

@Korijn Could you check if I interpreted them correctly? It was a bit tricky at times to work out what each function should do, considering that most don't have a docstring.


FirefoxMetzger:
TODO: some functions need renaming to adhere to the naming convention :)

pylinalg/func/vector.py
```diff
@@ -116,3 +116,114 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
     if dtype is not None:
         out = out.astype(dtype, copy=False)
     return out
+
+
+def project_camera(vector, projection_matrix) -> np.ndarray:
+    """Project a 3D vector into 2D space.
```
Korijn, line 122: looking at the implementation in pygfx, this is really just applying a matrix to a vector. I think it doesn't need a function here.

pylinalg/func/vector.py
```diff
@@ -116,3 +116,114 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+    projected_vector : ndarray
+        The projection of the vector.
+
+    """
+
+    raise NotImplementedError()
+
+
+def unproject_camera(vector, projection_matrix, /, *, depth=0) -> np.ndarray:
+    """Un-project a vector from 2D space to 3D space.
```
Korijn, line 142: unlike `project_camera` this function is more interesting than a simple matrix transformation, since there is the `depth` parameter, so this seems useful enough to keep around.

pylinalg/func/vector.py
```diff
@@ -116,3 +116,114 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+
+    Parameters
+    ----------
+    vector : ArrayLike
+        The vector to be un-projected.
+    projection_matrix: ArrayLike
+        The camera's intrinsic matrix.
+    depth : float
+        The distance of the unprojected vector from the camera.
```
Korijn, line 151: I'd say 0..1 would map from near to far?

pylinalg/func/vector.py
```diff
@@ -116,3 +116,114 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+
+    raise NotImplementedError()
+
+
+def from_matrix_position(homogeneous_matrix) -> np.ndarray:
+    """Return the position of the matrix (??)
+
+    @Korijn: I can't work out what this function does on a high-level. It
+    clearly extracts the bottom row of a homogeneous matrix, but why do we want
+    that?
```
Korijn, line 225: it just standardizes the conventions; where in the matrix is the translation component stored? this depends on whether the library is using row or column major matrix representations.

Korijn:
I went over the stubs and tried to provide additional info where applicable.

Of course all the conventions also still need to be applied (adding `out` and `dtype` keyword arguments, naming things, consistent docstrings, no typing, etc).

pylinalg/func/vector.py
FirefoxMetzger: I agree. I would actually cull it completely and not introduce an alias.

pylinalg/func/matrix.py
```diff
@@ -464,3 +464,31 @@ def matrix_make_orthographic(
...
+    (target).
+
+    Parameters
+    ----------
+    eye : ArrayLike
+        A vector indicating the direction that should be aligned.
+    target : ArrayLike
+        A vector indicating the direction to align on.
+    up : ArrayLike
+        @Korijn What does "up" do?
```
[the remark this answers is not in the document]
FirefoxMetzger: Ah, that makes sense. What's our coordinate frame convention? The z-axis is up? negative y is up?

pylinalg/func/vector.py
FirefoxMetzger: oh. I've never seen it stored in the bottom row nor considered it. Is that for performance reasons or pure convention?

pylinalg/func/matrix.py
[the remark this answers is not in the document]
almarklein: WGPU's NDC (normalized device coordinates), i.e. the coordinate system that you eventually render stuff in, defines positive y to be up, looking down z. https://wgpu-py.readthedocs.io/en/stable/guide.html#coordinate-system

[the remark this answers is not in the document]
almarklein: In PyGfx we don't adhere to a specific coordinate frame for the scene, because that differs per application anyway. But by default up is (0, 1, 0).

pylinalg/func/vector.py
Korijn: Um, could be either or both reasons! Depends on the data you're working with.

pylinalg/func/vector.py
Korijn: Me too!

pylinalg/func/vector.py
Korijn: Usually there are three matrices worth mentioning when rendering a 3D model to the screen.

World (or Model) matrix: transforms a model's vertices to world space.

View matrix: transforms world space coordinates to camera space. That means they're positioned "in front of" the camera, and "upright", with the camera at the origin.

Projection matrix: transforms view space coordinates to clip space/NDC (normalized device) coordinates. Typically there is some notion of a camera frustum which defines the transform. Coordinates that lie within the camera frustum end up on the [-1, 1] range on all three axes.

After this, the GPU takes over. Coordinates within the [-1, 1] clipping range are rendered onto the screen.

I probably got a bunch of details wrong but at least it gives you some handles to google for more info. ;)

FirefoxMetzger: > Usually there are three matrices worth mentioning when rendering a 3D model to the screen.

Ah okay. That indeed clarifies things; thanks! In robotics, we call the view matrix the _extrinsic matrix_ and the projection matrix the _intrinsic matrix_. Once we know the intrinsic matrix (through camera calibration), we can use reference points to estimate the extrinsic matrix and, once that is known, we can estimate World/Model matrices.

> I'd say 0..1 would map from the near to the far plane? Do you have a different idea?

I wouldn't limit the `depth` to 0-1. It would reduce the generality of the function and it's easy enough to pass `depth=z*(far - near)` to `unproject` and get the desired result.

almarklein: >  the [-1, 1] range on all three axes.

Worth mentioning that in OpenGL depth lies in  -1..1, but on wgpu it's 0..1.

> View matrix: transforms world space coordinates to camera space. That means they're positioned "in front of" the camera, and "upright", with the camera at the origin.

In PyGfx the view matrix is literally the inverse of the camera's object transform.

pylinalg/func/misc.py
```diff
@@ -0,0 +1,52 @@
...
+    -------
+    sphere : ndarray, [4]
+        A sphere (x, y, z, radius).
+
+    """
+
+    raise NotImplementedError()
+
+
+def abb_transform(aabb, homogenious_matrix, /, *, out=None, dtype=None):
```
FirefoxMetzger, line 29: @Korijn This one seems a little redundant as well, shall we cull it similarly to `vector_project`?

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+    -------
+    euclidean_coordinates : ArrayLike
+        A vector in euclidian coordinates.
+
+    """
+
+    raise NotImplementedError()
+
+
+def vector_distance_between(vectorA, vectorB, /, *, out=None, dtype=None):
```
FirefoxMetzger, line 207: @Korijn This function is `np.linalg.norm(vectorA-vectorB)`. Is it worth keeping?

FirefoxMetzger:
@Korijn Looking at the stubs, I've noticed that some of the functions can be written as one-liners in numpy. Is it worth keeping them around or is this "similar enough to not duplicate"?

FirefoxMetzger:
Also, CI doesn't seem to pick up this PR. Shall I look into the why, or are we not worried about this for now?

pylinalg/func/vector.py
almarklein: Maybe it's still worth for it's semantics - code that uses it becomes easier to read?

pylinalg/func/misc.py
Korijn: This one is not redundant. If the transform contains a rotation element, you cannot simply transform the two points expressing the extrema of the aabb. Instead, the method we employ [in pygfx](https://github.com/pygfx/pygfx/blob/f488647ac9a9f81f3e35ab0d72865a80d9308030/pygfx/linalg/utils.py#L36-L40) is to generate and transform all 8 corners of the aabb, and derive new extrema from them.

FirefoxMetzger: Ah, so the coordinate frame in which the bounding box is described doesn't transform together with the bounding box?

Then this function indeed has a lot of meaning. I'll update the docstring 👍

pylinalg/func/vector.py
FirefoxMetzger: You mean when you use `np.subtract` with `out` to perform the subtraction in place? That could work, but only if we know that we can overwrite one of the two input vectors.

I ran a profile where I compared `np.subtract` with `out` against the naive approach, and in-place is better for small arrays, but for large arrays `np.linalg.norm` appears to dominate runtime:

distance_timings

<details>
<summary>Code</summary>

```python
import numpy as np
import matplotlib.pyplot as plt
from timeit import timeit

rng = np.random.default_rng(42)
matrix_size = np.linspace(10, int(1e7), 100).astype(int)

vectorA = [
    rng.random((x, 3))
    for x in matrix_size
]

vectorB = [
    rng.random((x, 3))
    for x in matrix_size
]

n_repeats = 5

naive_times = []
optimized_times = []

for idx in range(len(matrix_size)):
    naive_times.append(
        timeit(
            "np.linalg.norm(vectorA[idx]-vectorB[idx], axis=-1)",
            globals=globals(),
            number=n_repeats,
        )
        / n_repeats
    )
    optimized_times.append(
        timeit(
            "np.linalg.norm(np.subtract(vector, vectorB[idx], out=vector), axis=-1)",
            setup="vector = vectorA[idx].copy()",
            globals=globals(),
            number=n_repeats,
        )
        / n_repeats
    )

plt.plot(matrix_size, naive_times, label="Naive")
plt.plot(matrix_size, optimized_times, label="Inplace")
plt.gca().set_title("Performance (log-log)")
plt.gca().set_xlabel("Number of Vectors")
plt.gca().set_ylabel("Time per run")
plt.xscale("log")
plt.yscale("log")
plt.legend(title="Method")
plt.savefig("distance_timings.png")

```

</details>

FirefoxMetzger:
@Korijn When you have time, could you look over the stubs once more and tell me if you agree with their docstrings? I'm happy with them as they are now and they should comply with the conventions of this repo. We also don't have any open discussions apart from the `distance_between` one, which we could move to a dedicated issue/PR for this function.

If you are happy with them, we could merge this in and I can start implementing them one-by-one :)

pylinalg/func/vector.py
Korijn: I still think it's useful to keep this around. Especially since we want it to broadcast when multiple vectors are passed.

pylinalg/func/matrix.py
```diff
@@ -464,3 +464,40 @@ def matrix_make_orthographic(
     out[3, 3] = 1
 
     return out
+
+
+def matrix_make_look_at(self, eye, target, up, /, *, out=None, dtype=None):
```
Korijn, line 469: I think you forgot to remove `self` here

pylinalg/func/misc.py
```diff
@@ -0,0 +1,59 @@
...
+    -------
+    aabb : ArrayLike
+        The transformed axis-aligned bounding box.
+
+    Notes
+    -----
+    This function preserves the alignment axes of the bounding box. This means
+    that the returned values are a new bounding box for the contained object
+    with respect to the previous alignment axes, not the new world coordinates
+    of the previous bounding box.
```
Korijn, line 56: It might be good to explicitly call out that this means the box will _grow_

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+    line), ``depth`` is used to select a single vector within.
+
+    Parameters
+    ----------
+    vector : ndarray, [2]
+        The vector to be un-projected.
+    projection_matrix: ndarray, [4, 4]
+        The camera's intrinsic matrix.
+    depth : number, optional
+        The distance of the unprojected vector from the camera.
```
Korijn, line 137: Is this the distance in world units, or a ratio of the distance between far and near? I'd personally prefer the latter, but either way, I'd like that to be expressed in the docstring

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+    Un-project a vector from 2D space to 3D space.
+
+    Find a ``vectorB`` in 3D euclidean space such that the projection
+    ``projection_matrix @ vectorB`` yields the provided vector (in 2D euclidean
+    space). Since the solution to the above is a 1D subspace of 3D space (a
+    line), ``depth`` is used to select a single vector within.
+
+    Parameters
+    ----------
+    vector : ndarray, [2]
```
Korijn, line 132: I think this should be `[3]`?

pylinalg/func/misc.py
```diff
@@ -0,0 +1,59 @@
...
+        A location into which the result is stored. If provided, it
+        must have a shape that the inputs broadcast to. If not provided or
+        None, a freshly-allocated array is returned. A tuple must have
+        length equal to the number of outputs.
+    dtype : data-type, optional
+        Overrides the data type of the result.
+
+    Returns
+    -------
+    aabb : ArrayLike
```
Korijn, line 48:
```suggestion
    aabb : ndarray, [2, 3]
```

pylinalg/func/misc.py
```diff
@@ -0,0 +1,59 @@
...
+    -------
+    sphere : ndarray, [4]
+        A sphere (x, y, z, radius).
+
+    """
+
+    raise NotImplementedError()
+
+
+def aabb_transform(aabb, homogenious_matrix, /, *, out=None, dtype=None):
```
Korijn, line 29: well, first of all, I think there's a spelling mistake here in the parameter `homogenious_matrix`, but also, I've called it `transform` or `matrix` elsewhere, I think we should be consistent where we can. my vote goes to `transform`, because I don't really know what `homogeneous` means in relation to a matrix

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
     if dtype is not None:
         out = out.astype(dtype, copy=False)
     return out
+
+
+def vector_unproject(vector, projection_matrix, /, *, depth=0, out=None, dtype=None):
```
Korijn, line 121: this function should broadcast over many vectors if possible, like the other functions in `vector.py` do

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+        A location into which the result is stored. If provided, it
+        must have a shape that the inputs broadcast to. If not provided or
+        None, a freshly-allocated array is returned. A tuple must have
+        length equal to the number of outputs.
+    dtype : data-type, optional
+        Overrides the data type of the result.
+
+    Returns
+    -------
+    rotated_vector : np.ndarray
```
Korijn, line 174:
```suggestion
    rotated_vector : ndarray, [3]
```

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+        A location into which the result is stored. If provided, it
+        must have a shape that the inputs broadcast to. If not provided or
+        None, a freshly-allocated array is returned. A tuple must have
+        length equal to the number of outputs.
+    dtype : data-type, optional
+        Overrides the data type of the result.
+
+    Returns
+    -------
+    euclidean_coordinates : ArrayLike
```
Korijn, line 199:
```suggestion
    euclidean : ndarray, [3]
```

pylinalg/func/vector.py
```diff
@@ -116,3 +116,195 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
...
+        A location into which the result is stored. If provided, it
+        must have a shape that the inputs broadcast to. If not provided or
+        None, a freshly-allocated array is returned. A tuple must have
+        length equal to the number of outputs.
+    dtype : data-type, optional
+        Overrides the data type of the result.
+
+    Returns
+    -------
+    distance : np.ndarray
```
Korijn, line 226:
```suggestion
    distance : ndarray
```

Korijn:
> Also, CI doesn't seem to pick up this PR. Shall I look into the why, or are we not worried about this for now?

Go ahead and look into it if you're going to be working on pylinalg, I would say it's going to be annoying otherwise. :)

> @Korijn When you have time, could you look over the stubs once more and tell me if you agree with their docstrings? I'm happy with them as they are now and they should comply with the conventions of this repo. We also don't have any open discussions apart from the `distance_between` one, which we could move to a dedicated issue/PR for this function.
> 
> If you are happy with them, we could merge this in and I can start implementing them one-by-one :)

I reviewed it again, found lots of minor nitpicks on the docstrings and a couple bigger questions. Happy to merge once this is addressed!

pylinalg/func/vector.py
FirefoxMetzger: Then we keep it 👍

pylinalg/func/vector.py
FirefoxMetzger: I might be wrong, but I don't think the camera matrix includes information on the distances of the far and near planes. We could add these as additional kwargs, but it might be cleaner to ask for distance in camera-space units (which may be different from world units if there is scaling between both).

pylinalg/func/matrix.py
```diff
@@ -464,3 +464,39 @@ def matrix_make_orthographic(
...
+
+
+def matrix_make_look_at(eye, target, up, /, *, out=None, dtype=None):
+    """
+    Rotation that aligns two vectors.
+
+    Given a camera (frame) looking in direction ``eye``, this function computes
+    a rotation matrix that makes the camera look into direction ``target``. It
+    then rotates the camera frame around ``target`` to make the camera's up
+    direction point in direction ``up``.
```
FirefoxMetzger, line 476: Having written this, I just realized that the signature for this function is incomplete. We don't know how a camera frame is oriented from a single vector  (`eye`), as the camera could have an arbitrary rotation around `eye`. 

Do we want to add a `old_up` argument to the function, or do we want to remove `up` and make the function perform a pure planar rotation from one vector onto another?

pylinalg/func/vector.py
FirefoxMetzger: Actually, the way this is implemented right now in pygfx both `project` and `unproject` are functions that are applied to matrices, but that live in `vector3.py`:

```python
    def unproject(self, camera) -> "Vector3":
        return self.apply_matrix4(camera.projection_matrix_inverse).apply_matrix4(
            camera.matrix_world
        )
```

Should we match that signature in `unproject` and move the function around?

FirefoxMetzger: The way this is implemented right now is that we take a 3-component vector in homogeneous space (u, v, 1) and project it into a 4-component vector in homogeneous space (x, y, z, 1).

@Korijn would you prefer to use homogeneous coordinates here, or would it be easier to have an interface that uses euclidean coordinates?

pylinalg/func/vector.py
FirefoxMetzger: @Korijn Sure. Should that work with only a single `matrix` (that is broadcasted to match the number of vectors) or with a batch of matrices, too.

pylinalg/func/matrix.py
Korijn: This line is incorrect: `Given a camera (frame) looking in direction ``eye``, this function computes a rotation matrix that makes the camera look into direction ``target``.`

The assumption is that the z-axis is the view axis, and you're going to create a rotation such that the z-axis aligns with the line from the point `eye` to the point `target`.

Refer to this implementation: https://github.com/mrdoob/three.js/blob/dev/src/math/Matrix4.js#L283-L328

For "up", we assume that the y-axis is the up axis, and you're going to rotate about the new view axis (from `eye` to `target`) such that the y-axis aligns with `up`.

FirefoxMetzger: Ah, so it's `(0, 0, 1) -> (target-eye)` and `(0, 1, 0) --> up`. I'll fix the docstring.

pylinalg/func/vector.py
Korijn: I'm having trouble explaining this well. Let me try a different angle. The typical use case for this function is to implement mouse picking, as done here for example: https://gamedev.stackexchange.com/a/12370/31164

pylinalg/func/vector.py
FirefoxMetzger: Then I'd say we keep this as is (`array, [2]`).

pylinalg/func/vector.py
FirefoxMetzger: Ah. The code you linked takes the screen coordinates of the mouse, normalizes them, then picks the corresponding _two_ points on the near and the far plane respectively, unprojects both, and then takes the difference. With the current signature it would look something like this for us:

```python
# normalized mouse coordinates
v, u = get_screen_pos_of_mouse() / (height, width)

near_camera = unproject((u, v), camera_matrix, depth=near)
far_camera = unproject((u, v), camera_matrix, depth=far)

result = far_camera - near_camera
result = result / np.linalg.norm(result)
```

This seems wasteful though because near_camera, far_camera, and the origin (0, 0, 0) always line up for a perspective camera. You'd only need one `unproject` and you can use any depth you desire. As such, you should also be able to do:

```python
# normalized mouse coordinates
v, u = get_screen_pos_of_mouse() / (height, width)
result = unproject((u, v), camera_matrix, depth=1)
```

(For orthographic cameras the problem is trivial as `unproject(anything, depth=value) = (0, 0, value)`, so users will not use `unproject`.)

The difference between the two approaches is that the length of the vector no longer matches the length of the viewing frustum measured along that ray; however, the code normalizes this vector anyway so that's likely unimportant :)

Btw, if the main use case is to cast a ray from the camera into the scene, we could add a `depth=None` default which will return a unit length vector, which saves the normalization step.

pylinalg/func/matrix.py
FirefoxMetzger: @Korijn Actually thinking about this again, could we drop `eye` in that signature and just do `target`?

`matrix_make_look_at(target, up, /, *, out=None, dtype=None)` sounds much more crisp, and we can get the current behavior via `matrix_make_look_at(target-eye, up)`, which I think is less surprising than passing in eye and target separately.

Korijn: Sure. You're then assuming eye is (0, 0, 0)?

FirefoxMetzger: No. At the moment the first step of `look_at` in pygfx is `_tmp_vector.sub_vectors(eye, target)`. I'm proposing to pass in `_tmp_vector` directly instead of passing in `eye` and `target`, because users can easily pass `look_at(target - eye, ...)` to get the previous functionality.

Since `tmp_vector` is an odd name in a signature, I would just name this `target`, because that is the actual direction the camera will look into as a result of the transformation.

Korijn: In that case I think the name `target` is misleading. Commonly lookat functions take a position `target`, not a direction. Let's either stick to the common signature, or clearly signal through different names that our implementation is not as usual.

FirefoxMetzger: I like `direction`; however, we can also keep it as-is if that means we match the common signature for this type of function. Your call.

Korijn: Personally, I'd prefer to keep `eye` and `target` as positions

FirefoxMetzger: Okay, then we keep it as it is right now :)

pylinalg/func/vector.py
FirefoxMetzger: > Is this the distance in world units, or a ratio of the distance between far and near? I'd personally prefer the latter, but either way, I'd like the unit to be expressed in the docstring for the depth parameter

And to answer the initial question here, the distance would be in world units.

FirefoxMetzger: @Korijn As far as I can see, this is the only conversation that needs resolving before this PR is ready. Are you okay with depth being in local coordinates instead of being a ratio (where 0 means on the near plane and 1 on the far plane)?

Korijn: Yes!

Korijn (approved):

FirefoxMetzger:
Interesting. I don't have write access to this repo. @almarklein @Korijn could one of you hit merge?

Korijn:
> Interesting. I don't have write access to this repo.

Fixed!

> @almarklein @Korijn could one of you hit merge?

Done!

PR #23, opened 2022-12-15
Fix CI not triggering on PRs

FirefoxMetzger:
In #22 The CI does not trigger when pushing commits to the branch. This PR aims to investigate why and then (hopefully) fix the issue.


FirefoxMetzger:
@Korijn RTR. Turns out we were triggering CI on the wrong branch (`master` instead of `main`).

Korijn (approved):

almarklein:
Classic 😄

Issue #21, opened 2022-11-25
Object API


Korijn:
Firstly, I'm confident that the current style, conventions and architecture for the functional API is right.

However, I'm still not completely sure on what would be right for the object oriented API.

I'm going to list some object oriented linalg/math APIs here for reference and inspiration:

* Unity
  * https://docs.unity3d.com/ScriptReference/Mathf.html
  * https://docs.unity3d.com/ScriptReference/Matrix4x4.html
  * https://docs.unity3d.com/ScriptReference/Transform.html
* pyquaternion 
  * https://github.com/KieranWynn/pyquaternion
  * http://kieranwynn.github.io/pyquaternion/
* three.js
  * https://threejs.org/docs/index.html#api/en/math/Matrix4
  * https://threejs.org/docs/index.html#api/en/core/Object3D
* Eigen
  * https://eigen.tuxfamily.org/index.php?title=Main_Page
  * https://eigen.tuxfamily.org/dox/GettingStarted.html
  * http://downloads.tuxfamily.org/eigen/eigen_plafrim_may_2011.pdf

FirefoxMetzger:
Allow me to add a C/C++ library as inspiration. It is very useful for computational geometry, which is the core of what we worry about here (unless I misunderstand). Maybe it can be another source for inspiration:

- CGAL 
    - Website: https://www.cgal.org/documentation.html
    - Packages: https://doc.cgal.org/latest/Manual/packages.html

FirefoxMetzger:
**TL:DR**: Proposal of an object-based API interface.

I've been thinking about a good design for this over the past week and I keep coming back to the same idea, so I want to propose it and see what you both think @Korijn @almarklein. It does something radically different from other existing libraries, but I think the difference is quite useful for us, especially because our focus really is on coordinate transformations (judging by the currently open issues).

In essence, I would not use `Matrix4x4`, `Point`, and `Vector` classes (which are glorified ndarrays anyway). Instead, I would use a `Frame` and a `Transform` class. We can still decide to introduce `Sphere`, `Capsule`, `Mesh`, and consorts, but we might want to have those in pygfx instead; happy to discuss :). The heavy lifting, however, is done by `Frame` and `Transform` (and its subclasses), which then allows us to do something like the following:

A (familiar?) example of a complex transformation. Given a point cloud on a unit sphere, find the points that face the camera:

<details>

```python
world = Euclidean3D()  # subclass of Frame

# define the location of a sphere
[...]

PR #25, opened 2022-12-18
Implement Object API

FirefoxMetzger:
Closes: #21 

This PR implements the object API discussed in #21.


FirefoxMetzger:
The tests are currently breaking due to precision issues when chaining two transforms. For example, when we have a transform with a large scaling and some translation then chaining it with its inverse will produce an identity matrix with "numerical zeros".

This currently fails numpy's default accuracy settings, i.e. the result isn't `rtol<=1e-7` and `atol=0` close to the expected identity. We could bump those thresholds, e.g., set `atol=1e-7` but this would mean that we will likely face issues with accumulating errors once we do longer chains of transformations.

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,154 @@
...
+        x, y, z, w = self._orientation
+        # fmt: off
+        result[:3, :3] = np.array([
+            [1 - 2*y**2 - 2*z**2,       2*x*y - 2*w*z,       2*x*z + 2*w*y],  # noqa: E201, E501
+            [      2*x*y + 2*w*z, 1 - 2*x**2 - 2*z**2,       2*y*w - 2*w*x],  # noqa: E201, E501
+            [      2*x*w - 2*w*y,       2*y*z + 2*w*x, 1 - 2*x**2 - 2*y**2],  # noqa: E201, E501
+        ]).T
+        # fmt: on
+
+        result[:, :3] = result[:, :3] * self._scale.reshape(1, 3)
```
Korijn, line 85: Is it better to use *= here?

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,154 @@
...
+        of this AffineTransform object. Typically, this maps from an object's
+        child frame into an object's parent frame.
+        """
+
+        result = np.zeros((4, 4))
+        result[-1, -1] = 1  # affine matrix :)
+
+        # credit to: http://www.songho.ca/opengl/gl_quaternion.htm
+        # @almarklein: Can we refactor quaternion_to_matrix to something like
+        # this? It's easier to read than the current implementation.
```
Korijn, line 75: I don't see why not.

Ideally the math in the object API is implemented using the functional API, to avoid diverging conventions and implementation details

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,154 @@
...
+            raise NotImplementedError()
+
+        other_rotation = qt.quaternion_to_matrix(other.orientation)[:3, :3]
+        position = other.position + (other.scale * self.position) @ other_rotation
+
+        orientation = qt.quaternion_multiply(self._orientation, other._orientation)
+
+        scale = self.scale * other.scale
+
+        return AffineTransform(position=position, orientation=orientation, scale=scale)
```
Korijn, line 130: I feel like this implementation (because it's not actually multiplying matrices) does not maintain the order of transformations. Is that right?

tests/conftest.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+from hypothesis.extra.numpy import arrays, from_dtype
+
+
+def normalize_quaternion(quaternion):
+    if np.linalg.norm(quaternion < 1e-16):
+        return np.array((0, 0, 0, 1))
+
+    return quaternion / np.linalg.norm(quaternion)
```
Korijn, line 9: Can you put these in the functional API?

pylinalg/obj/transform.py
FirefoxMetzger: sure!

pylinalg/obj/transform.py
FirefoxMetzger: It does maintain order :) (or at least it should ... if not there is a bug that needs fixing)

If you prefer, I could implement this as:

```python

new_transform = self.transformation_matrix @ other.transformation_matrix
position, orientation, scale = matrix_decompose(new_transform)
return AffineTransform(position=position, orientation=orientation, scale=scale)
```

The reason I didn't is that it is less accurate than working out the new `scale` and `orientation` manually. For the position, it's about the same. 

Also, we want to return a `AffineTransform` object to allow chaining statements like: `light_bulb.transform @ lamp.transform @ desk.transform @ floor.transform @ world.transform` to preserve the ability to modify/inspect the position, orientation, and scale, and to keep the possibility to optimize how we perform this computation in the future (to be more accurate, or faster, ...).

tests/conftest.py
FirefoxMetzger: Of course. I mainly wrote this to generate quaternions for testing via hypothesis, but we could turn this into a function that is available in the public API if you find it useful.

pylinalg/obj/transform.py
Korijn: Yes, it's perfectly fine! We have unit tests for every function, so. It should be good. :)

pylinalg/obj/transform.py
Korijn: I'm clearly not on your level math-wise, so I will just take it from you that order is maintained and accuracy is better this way. Cool!

tests/conftest.py
Korijn: I think you proved it's usefulness already :)

pylinalg/obj/transform.py
FirefoxMetzger: I'm happy to add more context. 

If we start with a classic affine matrix that includes rotation, scale, and translation then it would look something like this:

```python
scaleX*rot scaleY*rot scaleZ*rot posX
scaleX*rot scaleY*rot scaleZ*rot posY
scaleX*rot scaleY*rot scaleZ*rot posZ
0          0          0          1
```

We can segment the above into three separate matrices (`pos @ rot @ scale`). One that does the rotating, one that does translating, and one that does scaling:

```python
# scaling
scaleX 0      0      0
0      scaleY 0      0
0      0      scaleZ 0
0      0      0      1

# rotation
rot rot rot 0
rot rot rot 0
rot rot rot 0
0   0   0   1

# translation
1 0 0 posX
0 1 0 posY
0 0 1 posZ
0 0 0 1
```

The standard order in which we apply these matrices when going from child to parent is `child -> scale -> rot -> pos -> parent`. When chaining two affine transforms we simply build a longer sequence:

```
child -> scaleA -> rotA -> posA -> scaleB -> rotB -> posB -> parent
```

The million-dollar question now is: Can we reorder the terms in this sequence to bring this back to a familiar form? The answer is, ofc yes. If we respect the following rules:

```
rot -> scale <> scale -> rot  # order of scale and rotation doesn't matter
pos -> rot <> rot -> (rot @ pos)  # if I want to rotate first, I must translate with the rotated translation
pos -> scale <> scale -> (scale * pos)  # if I want to scale first, I must translate with the scaled translation

scale -> scale <> scale + scale  # two scales is the same as scaling once by the sum of scales
rot -> rot <> rot @ rot  # two rotations is the same as rotating by the combined rotation
pos -> pos <> pos + pos  # two translations is the same as translating by the sum
```

with that it becomes an exercise in simplification: all `scale`s to the left and all `pos` to the right:

```
child -> scaleA -> rotA -> posA -> scaleB -> rotB -> posB -> parent
child -> scaleA -> rotA -> scaleB -> (scaleB * posA) -> rotB -> posB -> parent
child -> scaleA -> scaleB -> rotA -> (scaleB * posA) -> rotB -> posB -> parent
child -> scaleA -> scaleB -> rotA -> rotB -> rotB @ (scaleB * posA) -> posB -> parent
child -> scaleA + scaleB -> rotA @ rotB -> rotB @ (scaleB * posA) + posB -> parent
```

This gives us our new scale, orientation, and position for `AffineMatrixA @ AffineMatrixB`:

```python
scale = scaleA + scaleB
orientation = rotA @ rot B = quaternion_multiply(orientationA, orientationB)
position = rotB @ (scaleB * positionA) + posB
```

Python calls this `self @ other`, so all that is left is substituting names.

Obviously, implementing just the last bit is faster than blowing up to matrices, multiplying, and then decomposing. What's more interesting, however, is that we also get higher numerical accuracy. If we track how each variable influences the result, it's easy to see that the new `scale` is unaffected by errors in `orientation` or position. Similarly, the new `orientation` is unaffected by any errors in `position` or `scale`. Only `position` "suffers" because we need to transform some of it as a result of the chaining.

tests/conftest.py
FirefoxMetzger: Okay. In this case, I would still keep a copy in `conftest.py`. The idea during a unit-test is to not trust the library you are testing, so it feels strange to use functions from that exact library to test the libraries behavior.

Korijn: Well. Think of the functional API as a separate package with its own tests (you'd have to add these). You are consuming that from the object API.

pylinalg/obj/transform.py
Korijn: I see. Thanks for elaborating, cool!

FirefoxMetzger:
I sank some time into figuring out how to write a good test for the inverse of an affine transform (i.e. how to reliably assert that `AffineTransform.inverse()` works correctly up to machine precision. I don*t have a full solution yet, but the current updates make some improvement on accuracy and on how the assertion is performed.

The current test does not check for the correctness of the computed positon (the troublemaker) and is currently incomplete. Will get there later once I have a good idea on how to test this.

FirefoxMetzger:
I just realized that the entire functional API actually right-multiplies vectors, i.e. we do `matrix @ vector` and not `vector @ matrix`. 

I will change this around in the `AffineTransform` class so that it matches the behavior of the functional API.

FirefoxMetzger:
Okay, this should now be up to date with the discussion in #21 . Feel free to have a look and let me know what you think.

poetry.lock
```diff
@@ -16,7 +16,7 @@ files = [
 dev = ["cloudpickle", "coverage[toml] (>=5.0.2)", "furo", "hypothesis", "mypy (>=0.900,!=0.940)", "pre-commit", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "sphinx", "sphinx-notfound-page", "zope.interface"]
```
Korijn, line 16: The lockfile should not be changed in this PR.

FirefoxMetzger: Good point. This is a residue from introducing hypothesis, which I did in this PR first, and then pulled into a separate PR once I realized that it would make sense to have available before we merge this here.

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class AffineTransform(Transform):
+    """
+    A affine transformation.
+
+    This transform represents an affine mapping between two coordinate frames:
+    parent and child. That is, a transformation that is linear in homogeneous
+    coordinates and that consists of a translation, rotation, and scale (applied
+    in that order).
```
Korijn, line 20: The order is always SRT (scale, rotate, translate)

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Parameters
+    ----------
+    position : ndarray, [3]
+        The position of the child's frame in the parent's space. (The
+        translation to apply to map vectors from child to parent.)
+    orientation : ndarray, [4]
+        The orientation of the parent's frame relative to the child's frame. (A
+        quaternion (format: x, y, z, w) describing the rotation from child to
+        parent.)
```
Korijn, line 31: Isn't it the other way around? When I set a child's rotation to 90 degrees about the y axis, I expect the child to rotate, not the parent.

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+    scale : ndarray, [3]
+        The scale of the parent expressed in child's units. (A axis-wise scaling
+        to map units from child to parent.)
```
Korijn, line 34: Here, too, I would expect this to be the other way around? Setting a scale of 2,2,2 on a child would make it twice as big, not its parent.

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+    def __init__(self, /, *, position=None, orientation=None, scale=None) -> None:
+        self._position = position
+        self._orientation = orientation
+        self._scale = scale
+
+        if position is None:
+            self._position = np.zeros(3)
+
+        if orientation is None:
+            self._orientation = np.array((0, 0, 0, 1))
```
Korijn, line 55: I think you may want to ensure this is an f8 ndarray like the other two, to avoid extra work casting.

Korijn:
I think this looks good enough as a first step and I think we could merge this as soon as the docstrings are made a little friendlier to readers with next to no math education.

That said, I had a conversation om this topic with Almar last friday and I do feel quite strongly that the parent/child mechanism should be implemented and tested here in (Affine)Transform. There is a lot to be said about it, so perhaps that is something we should get on a call for to discuss more efficiently than here on github. :)

pylinalg/obj/transform.py
FirefoxMetzger: Yes, I think we are using different semantics here. How about we avoid using `parent` and `child` while documenting this class? `parent` is aliased with the parent object in the scene graph, so it can become confusing when talking about the object's parent (object) vs the transform's parent (frame).

pylinalg/obj/transform.py
FirefoxMetzger: You are correct. I wrote the wrong documentation here. It should indeed be the orientation of the target frame when seen from the parent.

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+
+        result = qt.quaternion_to_matrix(self._orientation)
+        result[:, :3] *= self._scale.reshape(1, 3)
+        result[:3, -1] = self._position
+
+        return result
+
+    def inverse(self):
+        """
+        Invert this affine transformation.
```
almarklein, line 78: Avoid potential confusion of this being in-place.
```suggestion
        Get the inverse affine transformation.
```

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+    def orientation(self):
+        return self._orientation
+
+    @property
+    def scale(self):
+        return self._scale
+
+    @position.setter
+    def position(self, value):
+        self._position = np.asarray(value)
```
almarklein, line 129: Also cast to f8 if it's not already?

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+            raise NotImplementedError()
+
+        rotation = qt.quaternion_to_matrix(self.orientation)[:3, :3]
+        position = self.position + self.scale * (rotation @ other.position)
+
+        orientation = qt.quaternion_multiply(other._orientation, self._orientation)
+
+        scale = other.scale * self.scale
+
+        return AffineTransform(position=position, orientation=orientation, scale=scale)
```
almarklein, line 113: The current implementation assumes that the transform can always be described with a separate position, orientation, and scale. It is thus a subset of the affine transform, because it does not include shear. But this assumption breaks down when you compose two transforms, one of which has a rotation and the other a non-isotropic scale, since that will introduce shear.

This means that we can have multiple transforms, and set their properties as we please, but as soon as we compose two transforms, we don't know whether we need a matrix to describe it (instead of the properties).

 One way to solve this could be to produce a MatrixTransform when two transforms are composed. (or perhaps that should be called an AffineTransform, and this transform should have a different name, since it's really a subset of the Affine transform.)

Or, instead of properties it could have methods (`scale`, `rotate`, `translate`) and internally it would use the properties directly, or a matrix when it's needed.

Yet another solution could be to only allow non-isotropic scale at the leaf nodes?

pylinalg/obj/transform.py
```diff
@@ -0,0 +1,137 @@
...
+
+    """
+
+    # some constants for syntax sugar
+    FORWARD = np.array((0, 0, 1))
+    BACKWARD = np.array((0, 0, -1))
+    UP = np.array((0, 1, 0))
+    DOWN = np.array((0, -1, 0))
+    RIGHT = np.array((1, 0, 0))
+    LEFT = np.array((-1, 0, 0))
```
almarklein, line 44: I think that these should be props, that return the same values, except with the orientation applied.

pylinalg/obj/transform.py
almarklein: A variation of the first solution would be to check the case where shear is introduced, and only return a different `Transformation` type in that case.

FirefoxMetzger:
Just to give you a heads up on the pace. I'm currently visiting my family and only have access to a Chromebook, which isn't ideal for development. I will make changes to this PR after I return home, although I can participate in discussions in the mean time :)

pylinalg/obj/transform.py
Korijn: I would suggest to do this in a next PR?

almarklein: Yeah, that'd be fine.

pylinalg/obj/transform.py
FirefoxMetzger: Yeah, that's true. Not sure how I missed that. How about we add `AffineTransform.shear` and - for completeness - a `AffineTransform.projection` component? This way, we should be able to represent any affine transformation.

We can use a generic implementation of `AffineTransform @ AffineTransform` for now, where we construct two matrices internally, multiply them, and then decompose the result. Most of this is already implemented in the functional API, and all that is left is to update `matrix_decompose` to return a shear component as well. On top, we add a fast path for special cases where we know efficient solutions (e.g. multiplying with isotropic scale).

We would also have to decide in which coordinate system we want to specify each component. Currently, all parameters are specified in the source's frame (i.e., the parent frame), but we might want to choose something else depending on what you find intuitive.

pylinalg/obj/transform.py
FirefoxMetzger: Then I will leave this here until after our discussion meeting next week :)

FirefoxMetzger:
I updated the docstrings to be more concrete and also tried to phrase them in a more intuitive way. What's left is to decide how to deal with non-isotrophic scaling. (@almarklein )

almarklein:
So, having the ability to have non-uniform scaling can cause shear to be introduced, and we need to decide how to deal with that.

The current use-cases for having non-uniform scale are:  
- Flipping an object about an axis.  
- Flipping the camera (e.g. for 2D plots to make y point down, e.g. images).  
- Flipping text that is on a surface with a certain orientation. 
- Stretching objects, e.g. models or text.  
- Anisotropic images/volumes.  

For dealing with shear we have 3 options:  
- Do not allow non-uniform scaling.  Not really an option, as it's too useful.
- Only allow non-uniform scaling at the leaf nodes. Technically possible, but then we need to disallow objects with a non-uniform scale from having children which is annoying and confusing.
- Produce different type of transform when shear is included.
- Take shear into account in the AffineTransform class. 

From what I can see, Unity sort of does the last, so we should be able to do it too.

FirefoxMetzger:
@almarklein I'm done with the geometry submodule. If you like, feel free to review it.

Korijn:
> @almarklein I'm done with the geometry submodule. If you like, feel free to review it.

I guess your comment was supposed to go on https://github.com/pygfx/pygfx/pull/430 ?

FirefoxMetzger:
@Korijn Yep. My mistake. feel free to delete this one :)

FirefoxMetzger:
I will close this PR in favour of the discussions over at pygfx: https://github.com/pygfx/pygfx/pull/482

They include a transform module, which is essentially what we aimed to implement here. Once the PR is ready, we can choose if we wish to keep the transform module over at pygfx or if we want to hand it upstream and make it part of pylinalg.

PR #26, opened 2022-12-21
Add hypothesis as test dependency

FirefoxMetzger:
This PR introduces [hypothesis ](https://hypothesis.readthedocs.io/en/latest/index.html) as a new test dependency. 

Hypothesis is a package that allows property-based testing in python. This is like fuzzy testing (randomly generating inputs to tests to see if things break), but systematic in that property-based tests actively try to search for edge cases in the given search space. If a falsifying example is found, hypothesis simplifies it as much as possible to create a human-readable failure case that can be used for debugging and/or regression testing.

I've also refactored some of the current test cases to use hypothesis (as a demo on how to use it), and will use it to test the implemented stubs, as this is a much more reliable way to test code than testing with single examples.


FirefoxMetzger:
Lol. Looks like hypothesis may have found a bug in numpy 👀.

```python
# version A
versionA = np.eye(4)
versionA[:3, 3] = 2**32 + 1
versionA = versionA.astype(np.uint32)

# version B
versionB = np.eye(4, dtype=np.uint32)
versionB[:3, 3] = np.asarray(2**32 + 1)
```

The above snippet produces matching results on Windows, but differing results on Linux. The overflow is not handled correctly in `versionA`. I've [asked on SO](https://stackoverflow.com/questions/74872407/why-does-numpy-handle-overflows-inconsistently) about this, and if that turns out inconclusive I will write a ticked on numpy :)

Korijn:
Have you check the machine's native float precision? Emulation (such as on CI) can also have an effect.

FirefoxMetzger:
@Korijn Good thinking, but this isn't the cause, unfortunately. `2**32` is a python `int` that gets converted to a numpy/C `long`.

Korijn:
I mean, `versionA = np.eye(4)` creates a float64 array, right? So you're effectively converting 2**32+1 to an f8 and then back to uint32 with `astype`>

FirefoxMetzger:
> I mean, versionA = np.eye(4) creates a float64 array, right? So you're effectively converting 2**32+1 to an f8 and then back to uint32 with astype?

Ah, yes. I missed that completely. I'll check if that provides any insights.

---

@Korijn In the meantime, I think I will stop updating more tests and would like to get this reviewed/merged first. I did some modifications to the current functional API so it's a good point to stop, review, and continue with individual tests later.

First off, there is (I think) a bug in `matrix_make_rotation_from_euler_angles`. Rotations around the y-axis have the minus in the wrong position and thus rotate clockwise instead of counter-clockwise like the other axes.

Secondly, I might have misunderstood the desired behavior of `matrix_make_rotation_from_euler_angles`, but I assumed that the order in which of `angles` are given and the order of elements in the `order` kwarg matches. I.e., if I write `matrix_make_rotation_from_euler_angles([np.pi, 0 ,0], order="zyx")` then I get a 180° rotation around the z axis. However, currently this generates a 180° rotation around the x-axis. I've updated the function to apply angles with the given `order`, but later noticed that we have a unit test testing for the "faulty" behavior. Let me know if I am confusing this, and I will revert to the old behavior if desired.

FirefoxMetzger:
Changing the behavior of `matrix_make_rotation_from_euler_angles` breaks a bunch of unit tests in `quaternion_XXX`. @Korijn Let me know which behavior is desired (or what we consider correct) and I will either update the tests or the function accordingly.

Korijn:
> @Korijn In the meantime, I think I will stop updating more tests and would like to get this reviewed/merged first. I did some modifications to the current functional API so it's a good point to stop, review, and continue with individual tests later.

Got it.

> First off, there is (I think) a bug in `matrix_make_rotation_from_euler_angles`. Rotations around the y-axis have the minus in the wrong position and thus rotate clockwise instead of counter-clockwise like the other axes.

OK, nice catch!

> Secondly, I might have misunderstood the desired behavior of `matrix_make_rotation_from_euler_angles`, but I assumed that the order in which of `angles` are given and the order of elements in the `order` kwarg matches. I.e., if I write `matrix_make_rotation_from_euler_angles([np.pi, 0 ,0], order="zyx")` then I get a 180° rotation around the z axis. However, currently this generates a 180° rotation around the x-axis. I've updated the function to apply angles with the given `order`, but later noticed that we have a unit test testing for the "faulty" behavior. Let me know if I am confusing this, and I will revert to the old behavior if desired.

I modelled the function after [three.js' Euler object](https://threejs.org/docs/index.html?q=matrix#api/en/math/Euler). The "order" option is used to change the order in which the rotations are applied, not the order in which the arguments are passed to the function. So a call like in your example `[np.pi, 0 ,0], order="zyx"` is interpreted to mean a 180 degree rotation around the x-axis indeed. Essentially the `angles` array is always `[x, y, z]`.

FirefoxMetzger:
> The "order" option is used to change the order in which the rotations are applied, not the order in which the arguments are passed to the function.

Could we do both? Obviously, `order` should change the order in which rotations are applied, but I think that order should also match the order of `angles`. 

three.js sidesteps this by giving each rotation its own argument (`TREE.Euler(X, Y, Z, order)`). This way, it is clear that even if order changes, the first argument (`x`) marks rotation around the x-axis. Also, their approach has the limitation that not all possible euler-angles can be represented, e.g., the triplet `"yzy"` can not be expressed using tree's signature, because we can't set two values for `y`.

Korijn:
> Could we do both?

What do you propose? `matrix_make_rotation_from_euler_angles(x, y, z, order)`?

FirefoxMetzger:
> What do you propose? `matrix_make_rotation_from_euler_angles(x, y, z, order)`?

This PR updates the behavior to what I thought is intuitive. I.e., `matrix_make_rotation_from_euler_angles([np.pi, np.pi, 0],  order="zyx") rotates first by pi around the z-axis, then by pi around the y-axis, and then by 0 around the x axis.

We could take this one step further and allow all euler angles (currenty those that rotate twice around the same axis, e.g., "yzy", "zxz", ... are not supported).

Alternatively we could choose to match tree.js exactly  and do it following your suggestion.

Korijn:
> We could take this one step further and allow all euler angles (currenty those that rotate twice around the same axis, e.g., "yzy", "zxz", ... are not supported).

I don't really understand what this means. But if you think it's useful, go ahead.

For pygfx, I suppose users will want to set a rotation on a Transform via xyz euler angles, as it is the standard approach in Unity and Three. So as long as that is covered, I'm okay with your decision.

ivoflipse:
They could achieve "yzy" by applying 2 sets of rotations, but I can see how it would be nice if we allowed for the flexibility.

FirefoxMetzger:
> I don't really understand what this means. But if you think it's useful, go ahead.

There are 12 different conventions for how to apply euler angles, and we can use `order` to select which convention we would like to apply. `"xyz"` is one of these conventions, and because it is veey popular for game engines we make it the default.

With the current implementation, we can support 6 (50%) of these conventions right now and my suggestion was to add support for the remaining ones.
 
> For pygfx, I suppose users will want to set a rotation on a Transform via xyz euler angles, as it is the standard approach in Unity and Three. So as long as that is covered, I'm okay with your decision.

👍 

> They could achieve "yzy" by applying 2 sets of rotations, but I can see how it would be nice if we allowed for the flexibility.

True, we can call it 3 times (once for each axis) and then manually compose the result matrices to get any set of euler angles, even the extrinsic rotations that I havent mentioned yet. It is more error prone though, so it would be nice to have it ready-made and thoroughly tested, because its really easy to mess up the order of multiplications and end up with difficult to spot bugs...

Korijn:
Go for it!

almarklein:
I also appreciate the improved flexibility! However, changing the signature with respect to ThreeJS may confuse users familiar with ThreeJS. I think we should still go ahead, but let's put a explicit warning specifically for ThreeJS users in the docstring.

FirefoxMetzger:
Hm, if both of you are concerned about users migrating from tree.js what do you think about having the general `matrix_make_euler_rotation(angles, order="xyz", ...)` which works as we were discussing here and update the current function to match tree.js signature: `matrix_make_rotation_from_euler_angles(x, y, z, /,*, order=...)`?

Korijn:
Sure! It can re-use the other function and just provide a different signature.

almarklein:
Having two functions to do the same thing, but with a slightly different signature is confusing in its own way 😄. Let's go with what we think is the best API (your suggestion), but take care of our ThreeJS friends by clearly documenting that it's different from TreeJS.

almarklein:
🤣

Korijn:
I really don't care anymore this discussion is taking too long :) pick something and make it good!

FirefoxMetzger:
> I really don't care anymore this discussion is taking too long :) pick something and make it good! I already agreed to most of the things proposed here. I'm sure you can work something out

In that case, lets go with the current solution. One function that is general, plus a note for people coming from TreeJS. (TreeJS uses individual x, y, z, parameters, so people will stumble into that signature difference, check the docs, and then read about the behavior there.)

---

Otherwise, this should be RTR now :) 

I am currently thinking if it would make sense to move all the hypothesis strategies into a dedicated module or if we should keep them in conftest.py. Otherwise, I am happy with the PR.

pylinalg/func/matrix.py
```diff
@@ -134,36 +133,36 @@ def matrix_make_rotation_from_euler_angles(
     -------
     ndarray, [4, 4]
         Rotation matrix.
+
+
+    Notes
+    -----
+    If you are familiar with TreeJS note that this function uses ``order`` to
+    denote both, the order in which rotations are applied *and* the order in
```
almarklein, line 141: Nitpick: out-of-place comma
```suggestion
    denote both the order in which rotations are applied *and* the order in
```

Korijn (approved):

Korijn:
> I am currently thinking if it would make sense to move all the hypothesis strategies into a dedicated module or if we should keep them in conftest.py. Otherwise, I am happy with the PR.

Feel free to continue thinking about it during the next PR :)

PR #30, opened 2022-12-24
Add `quaternion_rotate` function

FirefoxMetzger:
This PR adds a function to the functional API that rotates a given vector using a quaternion without first converting it into a matrix. 

The current implementation is slower than its `matrix @ vector` variant, so I don't know if we want to add it in its current form or just do `matrix @ vector` internally. That said, I think it would be convenient to have this signature as part of the API :)

### DIFF
--- a/pylinalg/func/quaternion.py
+++ b/pylinalg/func/quaternion.py
@@ -215,6 +215,16 @@ def quaternion_rotate(vector, quaternion, /, *, out=None, dtype=None):
     dtype : data-type, optional
         Overrides the data type of the result.
 
+    Returns
+    -------
+    rotated_vector : ndarray, [3]
+        The input vector rotated by the given quaternion.
+
+    Notes
+    -----
+    For improved accuracy consider normalizing the vector before applying the
+    rotation and then re-apply the original scale afterwards.
+
     """
 
     vector = np.asarray(vector, dtype=dtype)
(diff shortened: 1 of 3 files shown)


Korijn (approved):

almarklein:
Out of curiosity, how come that it's slower? Wouldn't it be less multiplications overall?

Korijn:
> Out of curiosity, how come that it's slower? Wouldn't it be less multiplications overall?

If I can give any advice, it's to count not just the amount of math operations but also the amount of python language operations. An additional python function call is more expensive than an additional math operation if you can keep it local to C code with numpy.

FirefoxMetzger:
I agree with @Korijn. Its the number of times we cross the python/C barrier by making calls to numpy. This adds costant overhead which hits hard if we only process a single vector-quaternion pair.

That said if we were to vectorize this and apply it to a full list of vertices, we should be able to measure a speedup, because the the overhead armortizes.

almarklein:
Makes sense. Perhaps we should (eventually) implement both the direct and via-matrix approach and decide which one to take based on the number of vectors it is applied to ...

FirefoxMetzger:
Both are implemented now. The `quaternion_rotate` function introduced here does it directly (did not test vectorized application though) and the `quaternion_to_matrix() @ vector` approach allows you to use the matrix-based approach.

Issue #15, opened 2022-11-17
vector_spherical_to_euclidian


Korijn:
Convert spherical coordinates to euclidian

PR #35, opened 2023-02-03
implement vector_spherical_to_euclidean

FirefoxMetzger:
Closes: https://github.com/pygfx/pylinalg/issues/15

This PR implements the conversion from spherical coordinates to euclidean coordinates. It respects pygfx's convention for euclidean coordinates (y-axis up, negative z-axis forward). It also adds property-based tests to ensure that the transformation does the expected thing.

### DIFF
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -185,7 +185,8 @@ def vector_spherical_to_euclidean(spherical, /, *, out=None, dtype=None):
     Parameters
     ----------
     spherical : ndarray, [3]
-        A vector in spherical coordinates (r, phi, theta).
+        A vector in spherical coordinates (r, phi, theta). Phi and theta are
+        measured in radians.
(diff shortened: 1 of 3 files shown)


almarklein:
> It respects pygfx's convention for euclidean coordinates (y-axis up, negative z-axis forward). 

I think I'm guilty for stating that in pygfx cameras look down the negative z-axis by default, but I now think this should be positive z.

The NDC coordinates in wgpu define the depth-range to be `[0, 1]`, but it's up to the implementation to initialize depth to either 0 or 1, and then use a depth comparison function that is either `less`  or `greater`. In pygfx we initialize to 1 and use `less`- we look from 0 towards 1.

FirefoxMetzger:
Ah, so it should point along positive z. No problem. I'll update the PR.

Korijn (approved):

Korijn (approved):

Issue #16, opened 2022-11-17
vector_distance(a, b)


Korijn:
Distance between two vectors

PR #36, opened 2023-02-03
Implement vector_distance_between

FirefoxMetzger:
Closes: https://github.com/pygfx/pylinalg/issues/16

This PR implements `vector_distance_between`. Currently it hardcodes euclidean distance, though we could easily expose the other methods that numpy has. It also adds property-based tests and tests for raised exceptions.


almarklein (dismissed):

Korijn (approved):

FirefoxMetzger:
Odd. I can't reproduce the CI failure locally, so I triggered CI again and now it passes.

@Korijn My guess is that this was a failure specific to the hardware that GH used for that particular instance (since they are normalized, it might be an instance failure...). Anyhow, since I can't reproduce it I'd say we can afford to not worry about it for now and revisit it if it reoccurs in the future.

FirefoxMetzger:
LGTM from my side :)

FirefoxMetzger:
This _must_ be platform/CPU specific. The test effectively checks if the following two statements are equivalent:

```python
# note: vecA and vecB have shape (3,)
np.linalg.norm(vecA-vecB, axis=-1)
np.linalg.norm(vecA-vecB)
```

And sometimes they mismatch by a relative `2e-16`, which is very close to what 64-bit precision offers (`1e-16`). I also discovered that GH action runners are actually not normalized, i.e., under the hood runners may execute on different hardware, which matches the sporadic nature of the failure. If we hit the wrong CPU we get inaccurate results.

I think the solution for us is to simply allow a larger margin of error - we probably don't care for precision at that level - but damn that is an odd problem.

FirefoxMetzger:
Okay. I've created https://github.com/numpy/numpy/issues/23222 to track the issue in numpy ... let's see if this leads us anywhere.

In the meantime, we can just reduce the required precision we test for. Once the tests pass I am happy with this PR :)

Korijn (approved):

Issue #14, opened 2022-11-17
vector_euclidian_to_spherical


Korijn:
Convert euclidian coordinates to spherical coordinates

PR #39, opened 2023-02-06
implement vector_euclidean_to_spherical

FirefoxMetzger:
Closes: https://github.com/pygfx/pylinalg/issues/14

This PR implements `euclidean -> spherical` space transformation. It also adds a pb test; however, this will have to wait until https://github.com/pygfx/pylinalg/pull/35 is merged, because then we can add a round-trip test to ensure that the conversion does what we think it does.

### DIFF
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -279,7 +279,28 @@ def vector_euclidean_to_spherical(euclidean, /, *, out=None, dtype=None):
 
     """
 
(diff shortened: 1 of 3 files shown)


FirefoxMetzger:
@Korijn General question: Is there anything preventing us from using squash-merges instead of normal merges? I always find that squash-merging leads to a much cleaner git history.

Korijn (approved):

Issue #9, opened 2022-11-17
matrix_make_look_at


Korijn:
`matrix_make_look_at(eye, target, up, /, *, out=None, dtype=None)`

PR #40, opened 2023-02-09
implement matrix_look_at

FirefoxMetzger:
Closes: https://github.com/pygfx/pylinalg/issues/9

This PR adds `matrix_make_look_at` and an accompanying PB test. Interestingly this was not the most straightforward function to implement, so this one might require a review to ensure that my code is readable and simple enough.

### DIFF
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -485,8 +485,8 @@ def matrix_make_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
     In many cases, a natural choice for ``up_reference`` is the world frame's
     y-axis, i.e., ``up_reference`` would be the world's y-axis expressed in
     local coordinates. This can be thought of as "gravity pulling on the
-    rotation" (opposite direction of world frame's up) and creates a result with
-    a level attitude.
+    rotation" (opposite direction of world frame's up) and will create a result
+    with a level attitude.
     
 
     Parameters
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -1,6 +1,7 @@
 from math import cos, sin
 
 import numpy as np
+from numpy.lib.stride_tricks import as_strided
 
 
 def matrix_combine(matrices, /, *, out=None, dtype=None):
@@ -493,8 +494,41 @@ def matrix_make_look_at(eye, target, up, /, *, out=None, dtype=None):
     Returns
     -------
     rotation_matrix : ndarray, [4, 4]
-        A matrix describing the rotation.
+        A homogeneous matrix describing the rotation.
+
+    Notes
+    -----
+    The implementation ignores the scale of ``target - eye`` and ``up``, meaning
+    that the resulting rotation makes the coordinate frame axes _point_ in the
+    direction but won't scale the units to _match_ the direction.
 
     """
 
-    raise NotImplementedError()
+    eye = np.asarray(eye, dtype=float)
+    target = np.asarray(target, dtype=float)
+    up = np.asarray(up, dtype=float)
+
+    result_shape = np.broadcast_shapes(eye.shape, target.shape, up.shape)
+    if out is None:
+        out = np.zeros((*result_shape[:-1], 4, 4), dtype=dtype)
+    else:
+        out[:] = 0
+
+    # Note: The below is equivalent to np.fill_diagonal(out, 1, axes=(-2, -1)),
+    # i.e., treat the last two axes as a matrix and fill its diagonal with 1.
+    # Currently numpy doesn't support axes on fill_diagonal, so we do it
+    # ourselves to support input batches and mimic the `np.linalg` API.
+    n_matrices = np.prod(result_shape[:-1], dtype=int)
+    itemsize = out.itemsize
+    view = as_strided(out, shape=(n_matrices, 4), strides=(16 * itemsize, 5 * itemsize))
+    view[:] = 1
+
+    out[..., :-1, 2] = (target - eye) / np.linalg.norm(target - eye, axis=-1)
+    out[..., :-1, 1] = up / np.linalg.norm(up, axis=-1)
+
+    # Note: order is important to obtain a right-hand frame
+    out[..., :-1, 0] = np.cross(
+        out[..., :-1, 2], out[..., :-1, 1], axisa=-1, axisb=-1, axisc=-1
+    )
+
+    return out
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -470,7 +470,7 @@ def matrix_make_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
     """
     Rotation that aligns two vectors.
 
-    Given an entity at position `eye` looking in direction `target`, this
+    Given an entity at position `eye` looking at position `target`, this
     function computes a rotation matrix that makes the local frame "look at" the
     same direction, i.e., the matrix will rotate the local frame's z-axes
     (forward) to point in direction ``target - eye``.
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -481,13 +481,13 @@ def matrix_make_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
     (up) of a reference frame of choice expressed in local coordinates. The
     rotation around the new z-axis will then align `up_reference`, the new
     y-axis, and the new z-axis in the same plane.
-    
+
     In many cases, a natural choice for ``up_reference`` is the world frame's
     y-axis, i.e., ``up_reference`` would be the world's y-axis expressed in
     local coordinates. This can be thought of as "gravity pulling on the
     rotation" (opposite direction of world frame's up) and will create a result
     with a level attitude.
-    
+
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


FirefoxMetzger:
There is something funny about this function. If we allow arbitrary `eye`, `target`, and `up` vectors it may be impossible to construct a pure rotation matrix. This is because `target-eye` and `up` mustn't be orthogonal, but only if they are can we find a pure rotation that will map `z` onto `target-eye` and `y` onto `up`.

How do we want to handle these edge cases (especially considering that they might be "numerically orthogonal", i.e., close to orthogonal, but not exactly).

pylinalg/func/matrix.py
```diff
@@ -493,8 +494,41 @@ def matrix_make_look_at(eye, target, up, /, *, out=None, dtype=None):
...
     -------
     rotation_matrix : ndarray, [4, 4]
-        A matrix describing the rotation.
+        A homogeneous matrix describing the rotation.
+
+    Notes
+    -----
+    The implementation ignores the scale of ``target - eye`` and ``up``, meaning
+    that the resulting rotation makes the coordinate frame axes _point_ in the
+    direction but won't scale the units to _match_ the direction.
```
almarklein, line 503: > but won't scale the units to _match_ the direction.

What does this mean? If a user provides vectors that are not normalized, will it be the same as if they were normalized or not? Let's update this note so that this is clear.

FirefoxMetzger: It's more complicated then that, unfortunately. In the current implementation, the scale of `up` is ignored (though we want to change how `up` works anyway, so this is subject to change). We need to normalize `target-eye`, so its scale doesn't matter. The (relative) scale of `target` and `eye`, however, controls the direction of the new z-axis so these scales very much matter.

Hence the result will be different if all input vectors are normalized vs non-normalized. Maybe we just delete the note as it is more confusing than insightful?

FirefoxMetzger:
> I would say that we always want the result to be a pure rotation, i.e. not introduce skew. Instead we can use the up vector more as a hint than an "absolute" value.

@almarklein No problem. We do need to decide though what kind of hint it should be, i.e., how to compute the new y-axis from the inputs. A few approaches come to mind; however, none is universally "intuitive":

1. We fix the z-axis to point at `target-eye` and choose the y-axis as the unit vector that results from projecting `up` into the plane normal to the new z-axis. (We pick the closest value that keeps the new z-axis fixed.)
2. We find the vector that is normal to the target-eye-plane and use `up` to indicate the direction into which the new y-axis should point. (We pick the  closest value that keeps the target-eye-plane fixed.)
3. We do the same as in (1) and add a rotation around the new x-axis by half the angle between `up` and the new y-axis. (We pick the rotation that is closest to `eye`, `target`, `up` under MSE / euclidean distance)

All of these can be motivated, and as long as we document which one we choose we are probably fine. We could even add a `coorrection_mode` parameter to allow users to choose between them. At the same time, computing either correction is about as expensive as the entire `look_at` function without correction, so there is a performance aspect to consider if we "know" that `up` is already orthogonal to `target-eye`.

Korijn:
I would recommend to review the threejs implementation details for inspiration.

FirefoxMetzger:
Treejs uses the first option, i.e., it projects `up` into the plane orthogonal to the new z-axis and uses that vector to indicate the direction of the new y-axis.

I take it that we just want to copy what treejs is doing?

Korijn:
Yeah, that makes most sense to me.

However, what do you do when the up vector is exactly parallel to the z axis?

This commonly occurs with orbit camera controls where you can angle the camera upwards to align its viewing axis with the up/y axis.

almarklein:
> However, what do you do when the up vector is exactly parallel to the z axis?
>
> This commonly occurs with orbit camera controls where you can angle the camera upwards to align its viewing axis with the up/y axis.

My guess is there is nothing we can do except use an arbitrary up vector that is orthogonal to the z axis. For the orbit camera, this use-case can likely be prevented by limiting the elevation to between -89 and +89 degrees.

pylinalg/func/matrix.py
almarklein: Now I'm confused about what `target` and `eye` mean. I initially thought they were positions that would together define the new view direction.

FirefoxMetzger: Yes, thats what they do. However this is implemented as 

```python
new_z_axis = target - eye
new_z_axis /= np.linalg.norm(new_z_axis)
```

So the resulting vector is normalized and doesn't change the scale of an object. However, the actual direction is different if we use `target - eye` or `target - 2*eye` but the same if we use `target - eye` or `2*target - 2*eye`. Thus the (relative) scaling of `target` and `eye` matters.

FirefoxMetzger:
> However, what do you do when the up vector is exactly parallel to the z axis?

I was thinking about this problem in the gym this morning, and the reason we are having it is that using `(eye, target, up)` overconstrains the problem. Using just `(eye, target)`, we can get the new z-axis via `new_z=normalized(target-eye)`. This constrains 2 degrees of freedom; however, it doesn't specify the amount by which we should rotate around `new_z`.  In other words, we can still freely choose one angle (1 parameter to choose) while satisfying the constraint for `new_z`. We need to choose this parameter though to get a unique rotation.

Currently, we allow users to make this choice by offering them to choose a unit vector `up` (2 parameters), so we have to work out the missing rotation (1 parameter) from these two parameters (which can't work in all cases). To do this, we choose (1) from my proposals above, which is equivalent to constructing a plane that is orthogonal to `new_z - old_z` (wlog. the one that contains `new_z`), projecting both `old_y - new_z` and `up - new_z` into this plane, and finally measuring the angle from `old_y_projected` to `up_projected`. The result is a rotation that preserves `new_z` and that brings `old_y` as close as possible to `up` using only rotation. This, of course, breaks down if either `old_y_projected` or `up_projected` become 0, as we can no longer measure an angle between the two values.

An easy way to fix this would be to only allow users to specify `angle` (1 parameter) instead of `up` (2 parameters). In this case, there exists always exactly one rotation, but it breaks with ThreeJS's API (which, imho, is odd for this function anyway). 

I was also thinking of other ways that would seem like a "natural" fix, i.e., I was wondering: if I want to lerp from one orientation to another and use `look_at` to find intermediate matrices, is there a natural choice for these undefined spots that would result in a smooth lerp (assuming it goes through them). Unfortunately, the answer depends on the direction in which the lerp walks through these, so it's likely impossible to decide from the current inputs alone.

Korijn:
Isn't `angle` still defined against some orthogonal vector though? I'm not sure I understand how it would work.

Also here is threejs' workaround: https://github.com/mrdoob/three.js/blob/170323283ee44f99b41dc3d2825d92fe5e3d8300/src/math/Matrix4.js#L302

almarklein:
@FirefoxMetzger and I discussed this issue this morning. It looks like part of the confusion is because the API is geared towards object/camera related stuff, whereas pylinalg is more math oriented. For example, the `eye` and `target` are positions which could be replaced with a single arg that is actually a vector (e.g. `new_z`). Similarly `up` does not represent the new `y` asix, but defines what's considered up in the world space. BTW it seemed to help to think as  `up` as the direction of gravity (except pointing in the other direction), that pulls the "camera" straight.

Anyway, no matter how we name things, the semantics stay very specific to render engine stuff. Would it make sense to move this to pygfx instead?

FirefoxMetzger:
@Korijn What do you think of the updated version of the docstring of `look_at`? In that formulation, we could afford to not worry about the edge case of `new_z` and `up_reference` being aligned and delegate this to the user (since a good choice would be use-case dependent). 

> Isn't angle still defined against some orthogonal vector though? I'm not sure I understand how it would work.

The result is a rotation matrix, not a position/orientation, so unless I am mistaken we shouldn't need some orthogonal vector, just like I could say `a 90° rotation around the z-axis` and it would uniquely identify the rotation.

Korijn:
> @Korijn What do you think of the updated version of the docstring of `look_at`? In that formulation, we could afford to not worry about the edge case of `new_z` and `up_reference` being aligned and delegate this to the user (since a good choice would be use-case dependent).

This is great and very clear. I wonder if an exception is maybe better than a zero value? I imagine there may be valid inputs that could lead to a zero value as well?

almarklein:
>  I wonder if an exception is maybe better than a zero value? I dislike "silent failure".

Technically it's not a failure, since all rotations are simply equally close to the reference 😉  From a practically point of view, I think as a user I'd rather have weird rotations than a stack trace in a situation like that (or worse the app crashing).

FirefoxMetzger:
I've updated the behavior to only ever return a pure rotation.

Regarding the behavior in case of linearly dependent `z_new` and `up_reference`, I am in favor of @almarklein 's suggestion. Numpy's philosophy is to not do explicit error checking if it can be avoided and put the user in charge of ensuring that the inputs are sensible. This avoids expensive error checking in potentially critical paths, and also allows users to use functions in ways that may have been unforeseen at development time, but make sense for that use case (instead of artificially preventing this usage via an exception).

In our case, we could emit a warning if `z_new` and `up_reference` are dependent/close to being dependent. This check is a bit expensive (equivalent to matrix rank computation), but if you think it is worthwhile @Korijn we can add it.

Korijn:
No I think that sounds good. Nice work!

FirefoxMetzger:
In that case this is RTR.

pylinalg/func/matrix.py
```diff
@@ -465,13 +466,28 @@ def matrix_make_orthographic(
...
 
-def matrix_make_look_at(eye, target, up, /, *, out=None, dtype=None):
+def matrix_make_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
     """
     Rotation that aligns two vectors.
 
-    Computes a rotation matrix that rotates the input frame's z-axis (forward)
-    to point in direction ``target - eye`` and the input frame's y-axis (up) to
-    point in direction ``up``.
+    Given an entity at position `eye` looking in direction `target`, this
```
Korijn, line 473: Actually, `target` is also a position, not a direction.

Korijn (approved):

Issue #19, opened 2022-11-17
aabb_to_sphere


Korijn:
Convert an aabb to a sphere

Issue #20, opened 2022-11-17
aabb_apply_matrix


Korijn:
Apply an affine transformation matrix to an aabb

Korijn:
Closed by #41

PR #41, opened 2023-02-10
Implement both aabb ops

FirefoxMetzger:
Closes: https://github.com/pygfx/pylinalg/issues/19, https://github.com/pygfx/pylinalg/issues/20

This PR implements `aabb_transform` and `aabb_to_sphere` and adds PB-tests for each.

### DIFF
--- a/pylinalg/func/misc.py
+++ b/pylinalg/func/misc.py
@@ -1,5 +1,7 @@
 import numpy as np
 
+from .vector import vector_make_homogeneous
+
 
 def aabb_to_sphere(aabb, /, *, out=None, dtype=None):
     """A sphere that envelops an Axis-Aligned Bounding Box.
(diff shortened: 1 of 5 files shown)


pylinalg/func/misc.py
```diff
@@ -54,4 +67,14 @@ def aabb_transform(aabb, matrix, /, *, out=None, dtype=None):
...
-    raise NotImplementedError()
+
+    aabb = np.asarray(aabb, dtype=float)
+    matrix = np.asarray(matrix, dtype=float).transpose((-1, -2))
+
+    if out is None:
+        out = np.empty_like(aabb, dtype=dtype)
+
+    out[:] = (vector_make_homogeneous(aabb) @ matrix)[..., :-1]
+    out.sort(axis=-2)
```
Korijn, line 78: I don't think this is correct. You need to evaluate the transform on all 8 corners of the aabb to find the new extrema.

FirefoxMetzger: Yep, that's right. Not sure what I was thinking there. I'll update the code.

FirefoxMetzger: Done. @Korijn please check if you are okay with the heavy usage of `as_strided` here. I wanted to keep the ability to do batched `aabb_transform`, so I am being "clever" on how to build the array of corner points.

The result is efficient, but a bit verbose, so you need to decide if we want that here, or if we prefer something more readable but perhaps less efficient.

Korijn (approved):

PR #43, opened 2023-02-17
Update conventions

### LINKED ISSUE #24: Update conventions
We need to add a couple of things to the conventions document as became apparent in #22 

* Axis conventions (e.g. in look_at we assume positive z-axis is the view axis, positive y-axis is the up axis, etc)
* Which functions support broadcasting/vectorization, and how (e.g. most functions in vector.py support broadcasting a single transform such as a matrix or quaternion over many vectors)

### LINKED ISSUE #31: Document coordinate frame axes conventions
* Local space, view space, NDC space, wgpu space:
  * positive z is viewing direction
  * positive x is right
  * positive y is up

What is up in world space, depends on the application. This is why the camera and some routines in pylinalg have a configurable up vector.

Korijn:
This PR:

* Closes #24
* Closes #31


CONVENTIONS.md
```diff
@@ -29,6 +29,54 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+* World space to view space: view matrix
+* View space to NDC/clip space: projection matrix
+* NDC/clip space to screen space: screen matrix
+
+In pygfx, every world object has its own model matrix, which
+encapsulates scale, rotation and translation components. World objects
+are additionally organized hierarchically using parent/child relationships.
+The world matrix is then recursively defined for every world object; it is the
+result of multiplying an object's model matrix with its parent's world matrix.
+For the root world object, the world matrix is equal to its model matrix.
```
almarklein, line 51: I think this is incorrect. The model matrix defines the transform w.r.t. the parent, but for the root object (the scene) there is no root. In Pygfx this works out correctly - if the camera is placed in the scene - because the camera's world matrix (i.e. the inverse of the view matrix) contains the root model matrix.

CONVENTIONS.md
```diff
@@ -29,6 +29,54 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+operations in pygfx and applications based on pygfx, we adhere to a number of
+standard coordinate frame conventions to align with expectations from pygfx.
+
+Generally there are four transform steps when going from an object's local space,
+all the way to screen space:
+
+* Local space to world space: world matrix
+* World space to view space: view matrix
+* View space to NDC/clip space: projection matrix
+* NDC/clip space to screen space: screen matrix
```
almarklein, line 44: I think we can leave this out, since it's baked into the gpu?

Korijn: OK!

CONVENTIONS.md
Korijn: As far as I know, we use the inverse of the camera's world matrix as the view matrix.

It's not used as the root model matrix otherwise. We have plenty of examples where the camera is not a part of the scene, and where it is part of the scene, and it's not used as the root model matrix in any instance.

almarklein: Let me try to clarify. Imagine this scene:
```
scene  --> wobjec1 --> wobject2
(root)  \--> camera1

camera2
```

Semantically, `scene.matrix` should not be part of `wobject1.world_transform` or `wobject2.world_transform`, because the matrix defines the transform w.r.t. the parent, and since the scene is the root, it's own matrix should not be taken into account. In pygfx we wrongfully do take it into account. We should fix that.

Now when this scene is rendered with camera1, we actually get correct results, because `scene.matrix` is also included in the camera's world transform, which is inverted to get the view transform.

If the scene is rendered with camera2, and `scene.matrix` is not identity, we get wrong results.

almarklein: As for this text, I think "For the root world object, the world matrix is equal to its model matrix." should either be left out or something like "For the root world object, the world matrix is the identity transform, by definition."

Korijn: > Semantically, `scene.matrix` should not be part of `wobject1.world_transform` or `wobject2.world_transform`, because the matrix defines the transform w.r.t. the parent, and since the scene is the root, it's own matrix should not be taken into account. In pygfx we wrongfully do take it into account. We should fix that.

I respectfully disagree. Counter example:

```
         /--> tire
car     /
(root)  \--> camera1

camera2
```

The car has its own transform, in order to move around world space. That world space is an empty nothingness in this example is besides the point.

> If the scene is rendered with camera2, and `scene.matrix` is not identity, we get wrong results.

Do we? I don't think so?

almarklein: >>  If the scene is rendered with camera2, and scene.matrix is not identity, we get wrong results.
>
> Do we? I don't think so?

That depends on what you expect to happen 😆

almarklein: After a discussion, we can conclude that we can choose whatever we want. As it is written now, the reference frame is the "nothingness behind the root".

CONVENTIONS.md
```diff
@@ -29,6 +29,66 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+is configured independently on the camera and depends on its
+type (perspective, orthographic).
+
+Pylinalg provides methods in its functional API to produce all of these matrices.
+Since world matrices are of key interest to users and developers working with pygfx
+there need to be some conventions in place regarding its axes.
+
+* The positive Z axis is the forward direction.
+* The positive Y axis is the up direction.
+* The positive X axis is the right direction.
```
almarklein, line 64: Don't we need to decide on https://github.com/pygfx/pygfx/issues/460?

CONVENTIONS.md
```diff
@@ -29,6 +29,53 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+
+# Memory layout conventions
+
+Row-major can mean two things:
+
+* Memory layout; are rows or columns contiguous in memory
+* Are vectors columns or rows
+
+Numpy is row-major in both cases, and so we adhere to the same
+convention here in pylinalg.
```
FirefoxMetzger, line 91:
```suggestion
Pylinalg's kernels are written assuming a row-major (C-contigous) layout. If a
kernel supports batch processing of vetors, it assumes that the last dimension
contains the relevant vector data and that all other dimensions are batch
dimensions. As such, you can think of vectors being row vectors.
```

Actually, numpy doesn't have a notion of a vector being a row or column vector. 

The community just happens to stack data along the first axis by convention, which has its roots in good old SQL, where each row represents a data tuple. scikit-learn stuck to that convention, and now it's everywhere. Numpy terminology talks about `core` and `loop` dimensions for batched processing, but I don't think we need to introduce this notion.

FirefoxMetzger (commented):
Some suggestions on how we could improve these conventions

CONVENTIONS.md
```diff
@@ -144,40 +191,6 @@ it is not necessary to add them to pylinalg.
...
-* Names should be concise and short.
-* For every method, there is an alternative in-place method, signified
-  with the prefix `i`. This is not the prettiest option, but it is
-  concise and short.
-
-## Copying and mutability
-
-* By default, methods return new objects, and do not mutate self. Such
-  methods shall be referred to as "copying methods".
-* In-place methods return `self` to enable function chaining.
```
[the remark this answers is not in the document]
Korijn: This section was on the object api that is removed in this PR.

CONVENTIONS.md
```diff
@@ -29,6 +29,53 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+As the purpose of pylinalg as a library primarily is to support linear algebra
+operations in pygfx and applications based on pygfx, we adhere to a number of
+standard coordinate frame conventions to align with expectations from pygfx.
+
+Generally there are four transform steps when going from an object's local space,
+all the way to screen space:
+
+* Local space to world space: world matrix
+* World space to view space: view matrix
+* View space to NDC/clip space: projection matrix
```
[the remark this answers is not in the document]
Korijn: Well, this linalg library exists solely to support pygfx, it's not a general purpose linalg library. I do think it's relevant, we have a couple of functions in pylinalg that generate perspective and orthographic projection matrices, which map to NDC/clip space.

CONVENTIONS.md
```diff
@@ -29,6 +29,53 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
 
 All functions need to be covered by unit tests.
 
+# Coordinate frame conventions
+
+As the purpose of pylinalg as a library primarily is to support linear algebra
+operations in pygfx and applications based on pygfx, we adhere to a number of
+standard coordinate frame conventions to align with expectations from pygfx.
+
+Generally there are four transform steps when going from an object's local space,
```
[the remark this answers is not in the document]
Korijn: Thanks, very nice. I will include it.

CONVENTIONS.md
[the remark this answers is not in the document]
FirefoxMetzger: Oh, ofc. GH clipped the file just after `# Functional API conventions` and I didn't realize it. ignore this comment :D

CONVENTIONS.md
```diff
@@ -29,6 +29,53 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+encapsulates scale, rotation and translation components. World objects
+are additionally organized hierarchically using parent/child relationships.
+The world matrix is then recursively defined for every world object; it is the
+result of multiplying an object's model matrix with its parent's world matrix.
+For the root world object, the world matrix is equal to its model matrix. 
+
+The view and projection matrices are owned by camera objects. The view
+matrix can be acquired by inverting the camera's world matrix. The projection matrix
+is configured independently on the camera and depends on its
+type (perspective, orthographic).
```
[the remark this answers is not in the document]
Korijn: Yes, dependent in the sense that if you put the camera in the scene, it will influence the view matrix, indeed. That's correct.

CONVENTIONS.md
```diff
@@ -29,6 +29,53 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+
+Pylinalg provides methods in its functional API to produce all of these matrices.
+Since world matrices are of key interest to users and developers working with pygfx
+there need to be some conventions in place regarding its axes.
```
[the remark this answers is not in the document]
Korijn: Sure.

CONVENTIONS.md
```diff
@@ -29,6 +29,53 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+
+* The positive Z axis is the forward direction.
+* The positive Y axis is the up direction.
+* The positive X axis is the right direction.
+
+This means that if, for example, you add a car object to a scene, by default
+it is assumed that the car is facing towards positive Z. Due to the above axis
+conventions, you can then use `matrix_make_look_at` to easily create a new rotation
+matrix that makes the car look in another direction.
+
```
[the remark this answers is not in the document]
Korijn: Works for me!

CONVENTIONS.md
```diff
@@ -29,6 +29,85 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+all the way to screen space:
+
+* Local space to world space: world matrix
+* World space to view space: view matrix
+* View space to NDC/clip space: projection matrix
+
+The view and projection matrices are owned by camera objects. The view
+matrix can be acquired by inverting the camera's world matrix. The projection matrix
+is configured independently on the camera and depends on its
+type (perspective, orthographic).
```
FirefoxMetzger, line 78:
```suggestion
To render an object we have to express it's vertices in a camera's so-called NDC
coordinates, which stands for normalized device coordinates. Recall that
vertices are expressed in the object's *local* frame, which means we need to
work out the transformation from object *local* to NDC. We can do this by
following the chain of transformations in the scene graph from object *local*
via *world* to camera *local* and from camera *local* into NDC. Since this chain
is very important for rendering, it's parts have special names:

* **World Matrix**: The transformation from object's *local* into *world*.
* **View Matrix**: The transformation from *world* to camera's *local*. This is
  the inverse of the camera's world matrix.
* **Projection Matrix**: The transformation from camera's *local* into NDC.
```

We can introduce this by connecting to the frame conventions we used above. Also, having written this, would it flow better if this gets mentioned after the axis naming convention?

CONVENTIONS.md
almarklein: You seem to have commented earlier, but I had not seen your comment when I posted the above.

The disagreement here seems to be in the definition of the *root node* vs the *world space*. You say that they are the same. But it's equally valid to say that the world is the thing beyond the root node. It's a choice we have to make.

Personally, I initially felt that the former made more sense, but now I lean towards the latter. In part because in the case of  *root=world* we'd somehow need to enforce that the root node's transform is not used.

CONVENTIONS.md
```diff
@@ -29,6 +29,85 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+we use two additional frames to avoid confusion:
+
+* **source**: The source frame is the coordinate frame in which
+to-be-transformed vectors are expressed, i.e., it is the reference frame of the
+input vectors.
+* **target**: The target frame is the coordinate frame in which transformed
+  vectors are expressed, i.e., it is the reference frame of the output vectors.
+
+Generally there are four transform steps when going from an object's local space,
+all the way to screen space:
```
almarklein, line 69:
```suggestion
Generally there are three transform steps when going from an object's local space,
all the way to NDC:
```

Korijn:
So, here's where I think my personal confusion stems from. By default, these are the directions in three.js and also in pygfx:

image

* +X is right
* +Y is up
* -Z is forward

However, when you have a random object, and you use it's `lookAt` function to make it face a `target`, it's rotation will be adjusted such that it's local positive Z axis will point to `target`. See the screenshot below. I've translated the cylinder a little, and then made the cylinder "look at" the origin using `lookAt(0,0,0)`.

three-lookAt

FirefoxMetzger:
@Korijn Could this simply be a bug/inconsistency in ThreeJS?

almarklein:
I played with the editor too. Indeed, calling `lookAt` at an object turns it's local +Z axis toward the target. Doing the same with a camera turns it's -Z axis towards the target. --> Looks like the implementation of `lookAt` is different for camera's, which is an inconsistency. I wonder whether this has a historical thing that we can avoid by doing it right from the start, or whether there is another reason ...

Korijn:
> @Korijn Could this simply be a bug/inconsistency in ThreeJS?

No, I think it's intentional! There are similar differences when I try the Unity editor locally.

Korijn:
> --> Looks like the implementation of `lookAt` is different for camera's, which is an inconsistency. I wonder whether this has a historical thing that we can avoid by doing it right from the start, or whether there is another reason ...

Same for lights: https://github.com/mrdoob/three.js/blob/acb58df60448e75561c363f142f0df878705443d/src/core/Object3D.js#L284

Korijn:
I've updated the conventions document with your suggestions and included the Z-axis distinction for cameras and lights and other objects.

CONVENTIONS.md
```diff
@@ -29,6 +29,88 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+
+In addition to the above-mentioned reference frames, pylinalg uses a
+standardized naming scheme for the axes of a reference frame. The convention
+matches the convention chosen in pygfx:
+
+* The positive X axis indicates the right direction.
+* The positive Y axis indicates the up direction.
+* The Z axis is interpreted differently depending on the type of object:
+  * For cameras and lights, the _negative_ Z axis is the forward/viewing direction.
+  * For all other objects, the positive Z axis is the forward/viewing direction.
```
almarklein, line 76: Normal objects don't see.
```suggestion
  * For all other objects, the positive Z axis is the forward direction.
```

Korijn:
I think I've found the reasoning for three.js exception for cameras and lights, it is based on the glTF conventions.

* https://github.com/mrdoob/three.js/pull/14935
* https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#coordinate-system-and-units
* https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#view-matrix
* https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_lights_punctual#adding-light-instances-to-nodes

In the above links you can find the glTF conventions, including the exceptions to the Z-axis for cameras and lights.

So I think perhaps one of three.js goals is to be compatible with glTF assets out of the box. I wonder if that means their positive X axis points _left_ instead of _right_ in accordance with the glTF convention...

almarklein:
I'm still don't fully understand why the view direction of cameras and lights is the backward direction for normal objects. But I think we can assume that the ppl from glTF have thought about this well. So let's adopt this and also mention glTF instead of pygfx (and link to it too).

> glTF uses a right-handed coordinate system. glTF defines +Y as up, +Z as forward, and -X as right; the front of a glTF asset faces +Z.

This confused me, until I realized that if you rotate around these axiii, so that you view down the -Z axis, then +X is to the right again.

Korijn:
It also explains why the lookAt call in threejs just flips `eye` and `target` for cameras and lights.

Korijn:
@FirefoxMetzger would you mind taking over this PR? I think you are better at documenting the coordinate frames and conventions than I am, and also I fractured a rib yesterday so I won't be using my laptop a whole lot anytime soon (typing this on my phone).

FirefoxMetzger:
@Korijn Sure. I'll do a review and finish it up during the next sprint 👍 

> I fractured a rib yesterday

Ouch. Hope that things will heal back together smoothly.

Korijn:
I've split up this PR so we can merge #47 more quickly, and so that @FirefoxMetzger can take over this PR which is now isolated to just the changes to CONVENTIONS.md.

CONVENTIONS.md
```diff
@@ -29,6 +31,125 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+  to describe cameras. They use the following semantics:
+  * The positive X axis is called *right*.
+  * The positive Y axis is called *up*.
+  * The *negative* Z axis is called *forward*.
+* **Normalized Device Coordinates (NDC)** are represented using `(x, y, z)`
+  vectors and are used to describe points to render/plot. Points inside the unit
+  (half) box are rendered, others are not. NDC uses the following semantics:
+  * The positive X axis is called *right* and the box extent is `[-1, 1]`.
+  * The positive Y axis is called *up* and the box extent is `[-1, 1]`.
+  * The *negative* Z axis is the viewing direction and the box extent is `[0, 1]`.
```
almarklein, line 64: In NDC we look down the positive z-axis again, right?  An object at z=0.1 occludes an object at z=0.2.

almarklein (dismissed):
Looking good from my end! One minor comment.

CONVENTIONS.md
FirefoxMetzger: In that case NDC would be left-handed ... unless the negative x-axis is right?

Korijn: That's right, NDC is left handed.

CONVENTIONS.md
```diff
@@ -29,6 +31,125 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+right-handed coordinate frame. What differs depending on the type are the
+semantics used to describe each axis. Distances are measured in `m` (meters) and
+and angles in `rad` (radians) unless stated otherwise.
+
+* **Object Coordinates** are represented using `(x, y, z)` vectors and are used
+  to describe objects in a scene. They use the following semantics:
+  * The *negative* X axis is called *right*.
+  * The positive Y axis is called *up*.
+  * The positive Z axis is called *forward*.
+* **Camera Coordinates** are represented using `(x, y, z)` vectors and are used
```
Korijn, line 54: This also applies to Lights in glTF. See here: https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Khronos/KHR_lights_punctual#adding-light-instances-to-nodes

CONVENTIONS.md
FirefoxMetzger: Updated.

CONVENTIONS.md
FirefoxMetzger: Hm, I just thought that lights point along (0, 0, -Z) but are otherwise the same as normal objects. I.e., if I move a light "forward" by 2 units, will it travel along positive or negative Z?

Korijn: I think that's what is meant by

> the light's direction is defined as the 3-vector `(0.0, 0.0, -1.0)`

almarklein: > It's probably because shadow casting requires rendering the scene from the light's point of view, as if it were a camera.

Aha! 💡 

I think both camera and light shine/look forward, which is their `(0, 0, -1)` axis.

FirefoxMetzger: Then I will update to include lights as using camera coordinates.

CONVENTIONS.md
```diff
@@ -29,6 +31,129 @@ Black is left at defaults, flake8 and isort are configured to adhere to black.
...
+* **Spherical Coordinates** are represented using `(radius, theta, phi)` vectors
+  and use the following semantics:
+  * The *radius* measures the distance between a point and the origin and lies
+    between `[0, inf)`.
+  * The *phi* angle measures the counter-clockwise (CCW) rotation around the
+    positive y-axis. It is measured from the positive Z-axis and lies between
+    `[0, pi)`.
+  * The *theta* angle measures the counter-clockwise (CCW) rotation around the
+    negative X-axis. It is measured from the positive Y-axis and lies between
+    `[0, 2*pi)`.
```
FirefoxMetzger, line 75: I've updated the spherical coordinates to reflect the definition that the current implementation uses.

FirefoxMetzger:
If you both agree with the updates to the definition of spherical coordinates then I think we can merge. Looks good from my side otherwise :)

Korijn:
I'm good! Since I opened the PR, it requires you or Almar's approval.

almarklein (approved):

Korijn:
Nice work! This exercise will prove very much worthwhile.

PR #48, opened 2023-03-07
Add quaternion make from euler angles

FirefoxMetzger:
When I started porting pygfx yesterday I noticed we are missing the equivalent of `gfx.linalg.Quaternion().set_from_euler(...)`, which is used almost everywhere. (It's the thing that makes our examples rotate)

This PR adds this function under the name `quaternion_make_from_euler_angles`.

While implementing the tests, I've also found a bug that crept into some of the other functions related to rotation around cardinal axes: We rotate CCW around X and Z, but CW around Y. There is one function in our API that uses Euler angles: `matrix_make_rotation_from_euler_angles`, and after updating it to rotate CCW around Y, as per our conventions, a number of tests broke.

This leaves us with a choice: (1) We can update everything to rotate CCW around Y, or we can stick with rotating CW and document this difference. I prefer option (1), but can be flexible depending on how you guys think @Korijn @almarklein

### DIFF
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -153,6 +153,8 @@ def matrix_make_rotation_from_euler_angles(
         axis_idx = {"x": 0, "y": 1, "z": 2}[axis]
 
         matrix = np.array([[cos(angle), -sin(angle)], [sin(angle), cos(angle)]])
+        if axis_idx == 1:
+            matrix = matrix.T
         matrix = np.insert(matrix, axis_idx, 0, axis=0)
         matrix = np.insert(matrix, axis_idx, 0, axis=1)
         matrix[axis_idx, axis_idx] = 1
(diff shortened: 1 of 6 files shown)


Korijn:
Is this issue just local to pylinalg, or does the problem also exist in pygfx.linalg?

FirefoxMetzger:
> Is this issue just local to pylinalg, or does the problem also exist in pygfx.linalg?

I'm not sure yet. It took me until yesterday night to figure out what's wrong, and then I didn't have the energy to trace the impact. I will have a look today after I sorted out some bureaucracy.

FirefoxMetzger:
A quick test indicates that we are rotating CCW in pygfx when using euler angles:

```python
>>> import pygfx as fx
>>> import numpy as np
>>> angle = np.pi/2 
>>> quat = fx.linalg.Quaternion(0, np.sin(angle/2), 0, np.cos(angle/2))
>>> euler = fx.linalg.Euler(y=angle)
>>> fx.linalg.Quaternion().set_from_euler(euler) == quat
True
```

Korijn:
Then go ahead with option (1) :)

Korijn (approved):

PR #50, opened 2023-03-09
Cleaner namespace

almarklein:
* Use `__all__`.
* Add `version_info` to `__init__.py`.

### DIFF
--- a/pylinalg/__init__.py
+++ b/pylinalg/__init__.py
@@ -4,6 +4,9 @@ pylinalg
 Linear algebra utilities for Python.
 """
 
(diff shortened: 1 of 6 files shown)


Korijn (dismissed):
Nice!

Is `version_info` a "standard" like `__version__` is?

almarklein:
> Is version_info a "standard" like __version__ is?

Standard in the sense that many packages (including wgpu-py and pygfx :)) have it. Since it's a tuple you can actually do version comparisons on it, even when one of the version elements is larger than 9.

PR #54, opened 2023-03-14
make vector_apply_matrix broadcast inputs

FirefoxMetzger:
While porting pygfx I noticed that `vector_apply_matrix` can handle batches of vectors, but doesn't use broadcasting to do so limiting us to a single matrix. This PR rewrites the function to allow broadcasting of inputs (adhering to our convention that the last axes are core dimensions and everything else is a loop dimension).

### DIFF
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -95,9 +95,9 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
         transformed vectors
     """
 
-    vectors = np.asarray(vectors, dtype=float)    
+    vectors = np.asarray(vectors, dtype=float)
     matrix = np.asarray(matrix, dtype=float)
-    
+
     if out is None:
         out_shape = np.broadcast_shapes(vectors.shape[:-1], matrix.shape[:-2])
         out = np.empty((*out_shape, 3), dtype=dtype)
@@ -108,6 +108,7 @@ def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
     out[:] = result[..., :-1, 0]
 
     return out
+
 
 def vector_unproject(vector, matrix, /, *, depth=0, out=None, dtype=None):
     """
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -69,53 +69,44 @@ def vector_make_homogeneous(vectors, /, *, w=1, out=None, dtype=None):
 
 def vector_apply_matrix(vectors, matrix, /, *, w=1, out=None, dtype=None):
     """
-    Transform vectors by a transformation matrix.
+    Apply a transformation matrix to a vector.
 
     Parameters
     ----------
-    vectors : ndarray, [..., 3]
+    vectors : ndarray, [3]
         Array of vectors
     matrix : ndarray, [4, 4]
         Transformation matrix
-    w : number, optional, default 1
-        The value for the homogeneous dimensionality.
-        this affects the result of translation transforms. use 0 (vectors)
-        if the translation component should not be applied, 1 (positions)
-        otherwise.
+    w : ndarray, [1], optional
+        The value of the scale component of the homogeneous coordinate. This
+        affects the result of translation transforms. use 0 (vectors) if the
+        translation component should not be applied, 1 (positions) otherwise.
     out : ndarray, optional
-        A location into which the result is stored. If provided, it
-        must have a shape that the inputs broadcast to. If not provided or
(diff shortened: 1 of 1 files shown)


Korijn:
I'm not sure I fully appreciate the new possibilities. Can you give an example to illustrate?

almarklein:
IIUC, after this PR, the function can transform a batch of vectors with the same matrix, transform a vector with a batch of matrices, and both?

FirefoxMetzger:
@almarklein Exactly. Normally, SIMD kernels take a block of data (or multiple blocks with matching shapes) as input and perform the same operations element-wise on each element of the block (or on index-matched elements of multiple blocks). Kinda like a shader on the GPU. It so happens that SIMD can process multiple elements in one cycle (a mini-batch), so we actually walk over the data block(s) one mini-batch at a time and apply the same element-wise operations in parallel to each element of the mini-batch. GPU shaders do this, too, but abstract/hide the way they walk over the data.

Broadcasting is an extension of this idea, which allows (1) adding size-1 axes (to the left of an array/block) to match input dimensionalities, and (2) replicating elements along size-1 axes to match input shapes. (Note: this is a "virtual" operation that only affects metadata. We broadcast by creating a view into the actual data with updated strides and shape, but don't actually modify any data.)

In the case of `vector_apply_matrix`, the element-wise operation of the kernel consumes a vector of shape `(3,)` and a matrix of shape `(4, 4)` (numpy calls these _core dimensions_). All other dimensions indicate batches of data that the kernel walks over to apply the op element-wise (numpy calls these _loop dimensions_). Given a batch of vectors `shape=(batch, 3)` and a single `matrix` we need to broadcast `matrix` to `(batch, 4, 4)` to make the shapes match (prepend an axis and then replicate the matrix along that axis until it matches `batch`). This gives us inputs with matching shapes and allows numpy to do its thing.

The old kernel had the broadcasting of `matrix` hardcoded. If we _know_ that we always get a batch of vectors and a single matrix, then we can express the element-wise op as `np.dot(vectors, matrix.T)` . This is [equivalent](https://numpy.org/doc/stable/reference/generated/numpy.dot.html) to `vectors @ matrix.T`, but less scalable, because `np.dot` itself doesn't support broadcasting. The new kernel essentially just fixes this by using matmul instead, which makes the kernel more general and matches the behavior of numpy's kernels.

>  Can you give an example to illustrate?

```python
>>> import numpy as np
>>> import pylinalg as la
>>>
>>> matrices = np.empty((3, 4, 4))
>>> matrices[0] = la.matrix_make_translation((1, 2, 3))
>>> matrices[1] = la.matrix_make_translation((0, 0, 1))
>>> matrices[2] = la.matrix_make_translation((1, 0, 42))
>>>
>>> la.vector_apply_matrix((1, 0, 0), matrices)
array([[ 2.,  2.,  3.],
       [ 1.,  0.,  1.],
       [ 2.,  0., 42.]])
```

and, of course, operating on matrix-vector pairs element-wise:

```python
>>> vectors = np.eye(3)                        
>>> vectors
array([[1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]])
>>> la.vector_apply_matrix(vectors, matrices)   
array([[ 2.,  2.,  3.],
       [ 0.,  1.,  1.],
       [ 1.,  0., 43.]])
```

I don't think we have an explicit use case for multiple matrices or multiple batch dimensions yet, but I expect this to be useful when building apps. Say you want to place objects (text/glyphs?) along an arc. You need to work out the position along the arc, but you also need to rotate each object by some amount so that it's orientation aligns with the arc's normal:

```python

import numpy as np
import pylinalg as la

# quarter circle around origin in XY starting on the Y going CW
angles = np.linspace(np.pi/2, 0, 10)
radius = 5
positions = radius * np.stack([np.cos(angles), np.sin(angles), np.zeros_like(angles)])
rotations = la.quaternion_make_from_euler_angles(angles, order="Z")

# batch of transforms representing the local frames along the arc
# TODO: add broadcasting support to matrix_matke_transform
transforms = la.matrix_make_transform(positions, rotations, (0, 0, 0))

# glyph bounding box in local space (same for each glyph)
# anchored at (0, 0, 0)
glyph_aabb_local = np.array([
    [0, 0, 0],
    [1, 1, 0]
])

# batch of glyph bounding boxes in parent space
# result shape (10, 2, 3)
glpyh_aabb_object = la.vector_apply_matrix(glyph_aabb_local[None, ...], transforms[:, None, ...])

```

Korijn:
OK, cool! I'll approve when CI passes.

FirefoxMetzger:
LGTM.

Korijn (approved):

PR #55, opened 2023-03-15
allow broadcasting of simple rotations

FirefoxMetzger:
I noticed a bug in `la.quaternion_make_from_euler_angles`. If the user supplies a batch of `angles` but only rotates around a single axis (e.g. `order="x"`) then the current implementation produces an exception because it trips over the missing/implied core dimension of `angles`. This PR fixes that and makes batches of simple rotations behave like all the others.

### DIFF
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -146,7 +146,12 @@ def matrix_make_rotation_from_euler_angles(
     ccw around the y-axis, and finally 0° around the x axis.
 
     """
+    angles = np.asarray(angles, dtype=float)
     order = order.lower()
+
+    if angles.ndim == 0:
+        # add dimension to allow zip
+        angles = angles[None]
 
     matrices = []
     for angle, axis in zip(angles, order):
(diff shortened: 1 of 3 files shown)


almarklein:
Nice find! Maybe we should also add a test to prevent a regression.

I found the following approach work well:
* Define a list of individual items (e.g. angles, vectors), and a matching list of expected results.
* Test that the function works for these individually, in a for-loop.
* Then turn that list into an array, and apply the function to the batch, and test that result too.

FirefoxMetzger:
I'm always a bit cautious with loops inside a test, because it makes it difficult to set a breakpoint and jump to that line instantly while debugging (you end up stepping through the loop manually, hitting the breakpoint during each iteration).

My approach would have been to add a strategy to hypothesis that generates (broadcastable) batches of vectors and to test on these in addition to their flat counterparts.

almarklein:
Whatever works to test both the singular and batched usage 👍

Korijn:
> I'm always a bit cautious with loops inside a test, because it makes it difficult to set a breakpoint and jump to that line instantly while debugging (you end up stepping through the loop manually, hitting the breakpoint during each iteration).

Just as a tip if you find yourself in situations like that in the future, you can make breakpoints conditional on certain expressions.

Korijn (approved):

FirefoxMetzger:
Cool. I knew I can modify the code itself and set a breakpoint inside a branch, but it never occurred to me that you can set conditions on breakpoints in IDEs. Learned something new today.

PR #56, opened 2023-03-15
Broadcast matrix trafo

FirefoxMetzger:
This PR does the same update as the last 2 PRs (allow broadcasting); however this time for `matrix_make_transform` and all the functions it depends on.

### DIFF
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -31,14 +31,17 @@ def matrix_combine(matrices, /, *, out=None, dtype=None):
     ndarray, [4, 4]
         Combined transformation matrix.
     """
+
+    matrices = [np.asarray(matrix) for matrix in matrices]
+    result_shape = np.broadcast_shapes(*[matrix.shape for matrix in matrices])
+
     if out is None:
-        out = np.empty((4, 4), dtype=dtype)
-    out[:] = np.identity(4, dtype=dtype)
-    for matrix in matrices:
-        try:
-            np.dot(out, matrix, out=out)
-        except ValueError:
-            out[:] = np.dot(out, matrix)
+        out = np.empty(result_shape, dtype=dtype)
+
+    out[:] = matrices[0]
+    for matrix in matrices[1:]:
+        np.matmul(out, matrix, out=out)
+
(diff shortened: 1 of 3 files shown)


FirefoxMetzger:
This one requires a decision: In general we perform computations as 64-bit float. However, if `out` or `dtype` are given, then we accumulate the result in a buffer of a different type. At the same time, we ignore `dtype` if `out` is set. This can break (as it does here) if the dtype of `out` is not compatible with `float64`, e.g. if it is `int`.

Shall we raise in these cases (as the kernel currently does), or shall we convert to `out.dtype`? If the latter, when shall this conversion happen? Already during accumulation, or do we accumulate in a separate buffer and then convert if needed as the last step before the kernel returns. (This has implications on accuracy, since accumulating in a float32 is obviously less precise than doing so in float64.)

Korijn:
> Shall we raise in these cases (as the kernel currently does), or shall we convert to `out.dtype`? If the latter, when shall this conversion happen? Already during accumulation, or do we accumulate in a separate buffer and then convert if needed as the last step before the kernel returns. (This has implications on accuracy, since accumulating in a float32 is obviously less precise than doing so in float64.)

I would recommend mimicking the numpy convention, which is to raise an error:

```py
>>> out = np.empty((4,), dtype='f8')
>>> np.dot(np.eye(4, dtype=float), [1.,2.,3.,4.], out=out)
array([1., 2., 3., 4.])
>>> out = np.empty((4,), dtype='f4')
>>> np.dot(np.eye(4, dtype=float), [1.,2.,3.,4.], out=out)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 200, in dot
ValueError: output array is not acceptable (must have the right datatype, number of dimensions, and be a C-Array)
```

FirefoxMetzger:
That works, and since it is the expected thing to do, I don't think we need to explicitly document it in the conventions unless it comes up again as unclear.

It does mean, however, that the current test violates this convention (because it explicitly tests the opposite). I will delete that part of the test

Korijn (approved):

PR #59, opened 2023-04-05
(un)project along the z-axis

FirefoxMetzger:
I think I made a mistake when implementing the function to unproject. I don't remember the exact reason for why (I think it was at the time where we were discussing what exactly our frame conventions are), but unproject operates along the x-axis (first axis) instead of operating along the z-axis (third axis). This PR fixes that.


almarklein (approved):

FirefoxMetzger:
I've added a regression test to make sure that we are indeed projecting along the z-axis.

almarklein (approved):

Korijn (approved):

PR #60, opened 2023-04-05
Use offset when unprojecting

FirefoxMetzger:
Another bug in unproject ... when we use an asymmetric `depth_range` then we need to apply the position of the transformation matrix. Previously, we have ignored it.

@almarklein I also noticed some surprising behavior for `matrix_make_orthogonal` in how it handles the far and near plane. Could you have a look at the comment and see what makes sense for us?

### DIFF
--- a/pylinalg/func/matrix.py
+++ b/pylinalg/func/matrix.py
@@ -474,6 +474,16 @@ def matrix_make_orthographic(
     -------
     matrix : ndarray, [4, 4]
         orthographic projection matrix
+
+    Notes
+    -----
+    The parameters to this function are given in a left-handed frame that is
+    obtained by mirroring source's Z-axis at the origin. In other words, if the
+    returned matrix represents a camera's projection matrix then this function's
+    parameters are given in a frame that is like the camera's local frame except
+    that it's Z-axis is inverted. This means that positive values for `near` and
+    `far` refer to a negative Z values in camera local.
+
     """
 
     left = np.asarray(left, dtype=float)
--- a/tests/func/test_vectors.py
+++ b/tests/func/test_vectors.py
@@ -251,14 +251,25 @@ def test_vector_unproject(expected, projection_matrix):
 
 
 def test_unproject_explicitly():
+    # orthographic camera for a 100-unit block around origin
     matrix = la.matrix_make_orthographic(
-        250, -250, 250, -250, -100, 100, depth_range=(0, 1)
+        -100,  # left
+        100,  # right
+        100,  # top
+        -100,  # bottom
+        100,  # near (positive because camera faces along _negative_ z)
+        -100,  # far  (negative because camera faces along _negative_ z)
+        depth_range=(0, 1),
     )
-    expected = np.array(((250, 250, 0), (-250, -250, 0)))
+
+    # @almarklein: Is this really what we want? We set near=100 and far=-100, so
+    # shouldn't a value of local.z=100 correspond to NDC.z=0 (since it lives on
+    # the near plane) and local.z=-100 be NDC.z=1?
+    expected = np.array(((100, 100, -100), (-100, -100, 100)))
     projected = np.array(((1, 1), (-1, -1)))
 
-    actual = la.vector_unproject(projected, matrix)
-    assert np.allclose(actual, expected, rtol=1e-16, atol=np.inf)
+    actual = la.vector_unproject(projected, matrix, depth=(0, 1))
+    assert np.allclose(actual, expected)
 
 
 def test_vector_unproject_exceptions():
@@ -362,7 +373,7 @@ def test_vector_apply_matrix__perspective():
         assert vectors2[2][0] == cases[2][1]
 
 
-def test_vector_apply_matrix__orthographic():
+def test_vector_apply_matrix_orthographic():
     # Test for OpenGL, wgpu, and arbitrary depth ranges
     depth_ranges = (-1, 1), (0, 1), (-2, 9)
 
--- a/tests/func/test_vectors.py
+++ b/tests/func/test_vectors.py
@@ -251,25 +251,50 @@ def test_vector_unproject(expected, projection_matrix):
 
 
 def test_unproject_explicitly():
-    # orthographic camera for a 100-unit block around origin
-    matrix = la.matrix_make_orthographic(
-        -100,  # left
-        100,  # right
-        100,  # top
-        -100,  # bottom
-        100,  # near (positive because camera faces along _negative_ z)
-        -100,  # far  (negative because camera faces along _negative_ z)
-        depth_range=(0, 1),
+    # see https://github.com/pygfx/pylinalg/pull/60#discussion_r1159522602
+    # and the following comments
+
+    # cube at origin with side length 10
+    cube_corners = np.array(
+        [
+            [-5, -5, -5],
+            [5, -5, -5],
+            [5, 5, -5],
+            [-5, 5, -5],
+            [-5, -5, 5],
+            [5, -5, 5],
+            [5, 5, 5],
+            [-5, 5, 5],
+        ]
     )
+    cube_world_matrix = np.eye(4)
 
-    # @almarklein: Is this really what we want? We set near=100 and far=-100, so
-    # shouldn't a value of local.z=100 correspond to NDC.z=0 (since it lives on
-    # the near plane) and local.z=-100 be NDC.z=1?
-    expected = np.array(((100, 100, -100), (-100, -100, 100)))
-    projected = np.array(((1, 1), (-1, -1)))
+    # camera 10 units away from cube origin
+    camera_pos = (0, 0, 10)
+    cam_world_matrix = la.matrix_make_translation(camera_pos)
+    view_matrix = np.linalg.inv(cam_world_matrix)
 
-    actual = la.vector_unproject(projected, matrix, depth=(0, 1))
-    assert np.allclose(actual, expected)
+    # Scenario 1: near=4, far=16
+    projection_matrix = la.matrix_make_orthographic(
+        -10, 10, 10, -10, 4, 16, depth_range=(0, 1)
+    )
+    cube_local_to_cam_ndc = projection_matrix @ view_matrix @ cube_world_matrix
+    corners_ndc = la.vector_apply_matrix(cube_corners, cube_local_to_cam_ndc)
+    corner_in_view = np.all(
+        ((-1, -1, 0) < corners_ndc) & (corners_ndc < (1, 1, 1)), axis=-1
+    )
+    assert np.sum(corner_in_view) == 8
+
+    # Scenario 2: near=6, far=14
+    projection_matrix = la.matrix_make_orthographic(
+        -10, 10, 10, -10, 6, 14, depth_range=(0, 1)
+    )
+    cube_local_to_cam_ndc = projection_matrix @ view_matrix @ cube_world_matrix
+    corners_ndc = la.vector_apply_matrix(cube_corners, cube_local_to_cam_ndc)
+    corner_in_view = np.all(
+        ((-1, -1, 0) < corners_ndc) & (corners_ndc < (1, 1, 1)), axis=-1
+    )
+    assert np.sum(corner_in_view) == 0
 
 
 def test_vector_unproject_exceptions():
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -142,8 +142,8 @@ def vector_unproject(vector, matrix, /, *, depth=0, out=None, dtype=None):
 
     Notes
     -----
-    The source frame of this operation is the camera's local XY-plane and the
-    target frame is the camera's local frame.
+    The source frame of this operation is the XY-plane of the camera's NDC frame
+    and the target frame is the camera's local frame.
     """
 
     vector = np.asarray(vector, dtype=float)
@@ -161,7 +161,7 @@ def vector_unproject(vector, matrix, /, *, depth=0, out=None, dtype=None):
     vector_hom = np.empty((*vector.shape[:-1], 4), dtype=dtype)
     vector_hom[..., 2] = depth
     vector_hom[..., [0, 1]] = vector
-    vector_hom[..., 3] = 0
+    vector_hom[..., 3] = 1
 
     out[:] = (vector_hom @ inverse_projection.T)[..., :-1]
 
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -163,7 +163,9 @@ def vector_unproject(vector, matrix, /, *, depth=0, out=None, dtype=None):
     vector_hom[..., [0, 1]] = vector
     vector_hom[..., 3] = 1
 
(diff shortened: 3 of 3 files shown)


tests/func/test_vectors.py
```diff
@@ -251,14 +251,25 @@ def test_vector_unproject(expected, projection_matrix):
...
+        -100,  # bottom
+        100,  # near (positive because camera faces along _negative_ z)
+        -100,  # far  (negative because camera faces along _negative_ z)
+        depth_range=(0, 1),
     )
-    expected = np.array(((250, 250, 0), (-250, -250, 0)))
+
+    # @almarklein: Is this really what we want? We set near=100 and far=-100, so
+    # shouldn't a value of local.z=100 correspond to NDC.z=0 (since it lives on
+    # the near plane) and local.z=-100 be NDC.z=1?
```
almarklein, line 267: Yes, except that I'd expect `near < far`.

FirefoxMetzger: 😅 sorry for splitting hairs here, but I think its is really important to get exactly right: In the example above, where is the near plane? At `local.x=100` or at `local.x=-100`?

`near < far` is ambigous here, because we flip the sign of the z-axis when moving from local space to NDC space. The camdra's positive NDC.z axis points along the negative local.z axis, since cameras "look backwards". In which space do we define `near` and `far` when calling `matrix_make_orthogonal`? (An all the other frustum values for that matter.)

Korijn: I believe left/right/bottom/top/near/far are all expected to be expressed in a local input frame that has the following conventions:

-z near, +z far, -x left, +x right, -y bottom, +y top

I guess the function internally remaps?

FirefoxMetzger: > I believe left/right/bottom/top/near/far are all expected to be expressed in a local input frame

Maybe I am confusing myself here, but I don't think this is what is implemented right now. Here is the simplest example/test I can think of: 

An orthographic projection where we project the unit cube (local space) into the unit cube (NDC space). Expressed in local space, this means the near-plane lives at `near=1` and the far-plane at `far=-1`. For these inputs, the result should be an identity except for the Z-coordinate, which should be flipped (-1) since we switch from right-handed local to left-handed NDC, and cameras look down _negative_ Z. (To make the example even simpler, let's leave `depth_range=(-1,1)` for now.)

```python
>>> import pylinalg as la
>>> import numpy as np
>>> la.matrix_make_orthographic(-1, 1, 1, -1, 1, -1)
array([[ 1.,  0.,  0., -0.],
       [ 0.,  1.,  0., -0.],
       [-0., -0.,  1., -0.],
       [ 0.,  0.,  0.,  1.]])
>>> np.allclose(la.matrix_make_orthographic(-1, 1, 1, -1, 1, -1), np.eye(4))
True
>>> # coordinates on the near-plane
>>> la.vector_unproject((0, 0), projection, depth=-1) 
array([ 0.,  0., -1.])
>>> # coordinates on the far plane
>>> la.vector_unproject((0, 0), projection, depth=1)  
array([0., 0., 1.])
```

This would suggest that the function is currently not flipping. Instead, we manually add the flip by swapping the values passed to `near` and `far`:

```python
>>> projection = la.matrix_make_orthographic(-1, 1, 1, -1, -1, 1)  
>>> projection
array([[ 1.,  0.,  0., -0.],
       [ 0.,  1.,  0., -0.],
       [ 0.,  0., -1.,  0.],
       [ 0.,  0.,  0.,  1.]])
>>> la.vector_unproject((0, 0), projection, depth=-1)
array([0., 0., 1.])
>>> la.vector_unproject((0, 0), projection, depth=1)  
array([ 0.,  0., -1.])
```

This has the "odd" consequence that `depth=-1` (near-plane, recall: `depth_range=(-1, 1)`) we get the Z value we pass to `far` and for `depth=1` (far-plane) we get the value we pass to `near`.

@Korijn Are we happy with the behavior as it is now, or should we add a flip to the function such that `near` corresponds to `depth_range[0]` and `far`  corresponds to `depth_range[1]`?

FirefoxMetzger: >[...] by placing each side a certain distance from the origin. This image illustrates well how the input arguments are used:

@Korijn In response to your edit: I get the idea (I think) but the image is a bit confusing. The origin is always the tip of the frustum and as a result, the distance between the origin and the top, bottom, left, and right plane of the frustum is always 0 😅.

Korijn: I can give you a use scenario with which you can validate the implementation details:

* I put a camera 10 units away from a cube with sides of length 10.
* The camera is looking straight at the cube's center.
* If I configure the near and far planes at 4 and 16, I can see the cube on screen (since the camera frustum is just wide enough to envelop the cube).
* If I configure the near and far planes at 6 and 14, I can't see the cube on screen (since I'm only viewing the inside of the cube).

I would expect to be able to pass the near and far plane values to the camera exactly as written, so:

```py
la.matrix_make_orthographic(left, right, top, bottom, 4, 16)  
```

It doesn't really matter how the camera and cube are positioned and oriented.

almarklein: > The origin is always the tip of the frustum and as a result, the distance between the origin and the top, bottom, left, and right plane of the frustum is always 0 😅.

The left, right, top, bottom apply at the near plane 😉

almarklein: Oh, but we're talking about an ortho transform here, so the frustrum is a cuboid, not a pyramid.

FirefoxMetzger: > I can give you a use scenario with which you can validate the implementation details:

Here's my mock-up of the scenario(s) in pure pylinalg:

```python
import numpy as np
import pylinalg as la

# cube at origin with side length 10
cube_pos = (0, 0, 0)
cube_corners = np.array([
    [-5, -5, -5],
    [ 5, -5, -5],
    [ 5,  5, -5],
    [-5,  5, -5],
    [-5, -5,  5],
    [ 5, -5,  5],
    [ 5,  5,  5],
    [-5,  5,  5],
])
cube_world_matrix = np.eye(4)

# camera 10 units away from cube origin
camera_pos = (0, 0, 10)
cam_world_matrix = la.matrix_make_translation(camera_pos)
view_matrix = np.linalg.inv(cam_world_matrix)

# Scenario 1: near=4, far=16
# --------------------------
projection_matrix = la.matrix_make_orthographic(-10, 10, 10, -10, 4, 16, depth_range=(0, 1))

# cube corners in NDC
cube_local_to_cam_ndc = projection_matrix @ view_matrix @ cube_world_matrix
corners_ndc = la.vector_apply_matrix(cube_corners, cube_local_to_cam_ndc)

corner_in_view = np.all(((-1, -1, 0) < corners_ndc) & (corners_ndc < (1, 1, 1)), axis=-1)
print(f"(Scenario 1) Corners in view: {np.sum(corner_in_view)}")
# (Scenario 1) Corners in view: 8

# Scenario 2: near=6, far=14
# --------------------------
projection_matrix = la.matrix_make_orthographic(-10, 10, 10, -10, 6, 14, depth_range=(0, 1))

# cube corners in NDC
cube_local_to_cam_ndc = projection_matrix @ view_matrix @ cube_world_matrix
corners_ndc = la.vector_apply_matrix(cube_corners, cube_local_to_cam_ndc)

corner_in_view = np.all(((-1, -1, 0) < corners_ndc) & (corners_ndc < (1, 1, 1)), axis=-1)
print(f"(Scenario 2) Corners in view: {np.sum(corner_in_view)}")
# (Scenario 2) Corners in view: 0
```

They indeed behave as you intend in that the first scenario will have all 8 corners in view, whereas the second has no corners in view.

The problem isn't so much that, but rather how the values (`near=4`, `far=16`) relate to our (coordinate) convention. Cameras use camera coordinates in which _negative_ Z is forward. In other words, the cube's corners expressed in the camera's local frame are (note the negative z-component):

```python
# cube corners in camera's local space
print(la.vector_apply_matrix(cube_corners, view_matrix @ cube_world_matrix))
# [[ -5.  -5. -15.]
#  [  5.  -5. -15.]
#  [  5.   5. -15.]
#  [ -5.   5. -15.]
#  [ -5.  -5.  -5.]
#  [  5.  -5.  -5.]
#  [  5.   5.  -5.]
#  [ -5.   5.  -5.]]
```

If we are strict with our convention and `la.matrix_make_orthographic` takes values in the camera's local frame, then `near=4` and `far=16` should not see the cube and we should have to pass `near=-4` and `far=-16` instead to see the cube.

What I'm wondering is if we should keep the current behavior (and maybe update our convention), or if we should update the behavior.

almarklein: > This has the "odd" consequence that for depth=-1 (near-plane, recall: depth_range=(-1, 1)) we get the Z value we pass to far and for depth=1 (far-plane) we get the value we pass to near.

I think its clearer if simply place the points to project in front of the camera (making their `z` negative). Otherwise this gets weird if we apply it to perspective projection. Let's also use easier to recognize values for near and far:

```py
>>> projection = la.matrix_make_perspective(-1, 1, 1, -1, 0.5, 5, depth_range=(0, 1))
...
>>> # Near plane
>>> x=projection @ (0, 0, -0.5, 1); x[:3]/x[3]
array([0., 0., 0.])
>>> la.vector_apply_matrix((0, 0, -0.5), projection)
array([0., 0., 0.])
>>> # Far plane
>>> x=projection @ (0, 0, -5, 1); x[:3]/x[3]
array([0., 0., 1.])
>>> la.vector_apply_matrix((0, 0, -5), projection)
array([0., 0., 1.])
```
All looks well, now for unproject:
```py
>>> # Near plane
>>> x=np.linalg.inv(projection) @ (0, 0, 0, 1); x[:3]/x[3]
array([ 0. ,  0. , -0.5])  # ok
>>> la.vector_unproject((0, 0), projection, depth=0)
array([0., 0., 0.])  # wrong
>>> Far plane
>>> x=np.linalg.inv(projection) @ (0, 0, 1, 1); x[:3]/x[3]
array([ 0. ,  0. , -5.])  # ok
>>> la.vector_unproject((0, 0), projection, depth=1)
array([0., 0., 0.])  # wrong
```

I ran this with the current `main`, this PR should make that unproject matches the result of the raw matrix multiplication.

almarklein: > If we are strict with our convention and la.matrix_make_orthographic takes values in the camera's local frame, then near=4 and far=16 should not see the cube and we should have to pass near=-4 and far=-16 instead to see the cube.

The `near` and `far` are defined as distances from the camera, with negative distance meaning behind the camera. So they're not "in a particular coordinate frame", just distances (albeit signed).

Korijn: > Here's my mock-up of the scenario(s) in pure pylinalg:
> 
> <code>
> 
> They indeed behave as you intend in that the first scenario will have all 8 corners in view, whereas the second has no corners in view.

This is exactly right! And so is Almar's code example.

> What I'm wondering is if we should keep the current behavior (and maybe update our convention), or if we should update the behavior.

Keep the current behavior for `matrix_make_orthographic` and `matrix_make_perspective`. These are implemented correctly.

> this PR should make that unproject matches the result of the raw matrix multiplication.

Exactly.

> and maybe update our convention

What do you propose?

FirefoxMetzger: > ```
> >>> la.vector_unproject((0, 0), projection, depth=0)
> array([0., 0., 0.])  # wrong
> ```

@almarklein Yes, this PR fixes these problems, which come from `depth_range` not being symmetrical.

> The near and far are defined as distances from the camera, with negative distance meaning behind the camera. So they're not "in a particular coordinate frame", just distances (albeit signed).

Yes, this is the part that was confusing me. `left`, `right`, `top`, and `bottom` look like they are given in local space, but `near` and `far` are not. Instead, they are given in some other frame that we have not (yet?) formally introduced. This new frame is a step between local and NDC. It is like local except that its Z-axis is flipped, and is like NDC except that no shear or scale is applied yet.

> What do you propose?

I would document that the parameters for `matrix_make_orthographic` are given in this in-between space (like local but with flipped Z). Once this is clear, everything falls back into place, but if you think parameters are in local space you go down a rabbit hole of confusion 😄

FirefoxMetzger:
Okay, I made a proposal for the docs in this PR based on our discussion (thanks for the clarification @almarklein @Korijn ). I've also updated the explicit unit test to use the example that @Korijn made earlier, which I think gives it more context than the generic one that I had there before.

I think I'm happy with this one and will address future points (if any) in upcoming PRs.

almarklein:
> I made a proposal for the docs in this PR based on our discussion 

I'm not sure if introducing another coordinate frame will avoid confusion. The args are distances, not points that need to live somewhere. Perhaps it suffices to add a small note to the `near` and `far` plane arg docs, that a positive distance is defined as being in the +z direction?

pylinalg/func/matrix.py
```diff
@@ -474,6 +474,16 @@ def matrix_make_orthographic(
...
         orthographic projection matrix
+
+    Notes
+    -----
+    The parameters to this function are given in a left-handed frame that is
+    obtained by mirroring source's Z-axis at the origin. In other words, if the
+    returned matrix represents a camera's projection matrix then this function's
+    parameters are given in a frame that is like the camera's local frame except
+    that it's Z-axis is inverted. This means that positive values for `near` and
+    `far` refer to a negative Z values in camera local.
```
Korijn, line 485: you may want to copy this or refer to this in `matrix_make_perspective`

FirefoxMetzger:
> The args are distances, not points that need to live somewhere.

@almarklein But parameters can be negative, and both `left` and `bottom` typically are. Strictly speaking, distances can never be negative, so we are actually not specifying distances, but rather points along the respective coordinate axes.

> I'm not sure if introducing another coordinate frame will avoid confusion.

I agree that it adds complexity to the docs, but I think it is mostly because we are now pointing out the complexity that is inherent in the function rather than us adding complexity artificially.

The problem I'm trying to solve is this: *For debugging, I wanted to create an orthographic camera that has a frustum that matched a particular box in world space. What values do I need to plug into `matrix_make_orthographic` to get such a projection?* I checked the docs and found no explicit mention of the reference frame for parameters. Next, I inspected places where it was used and it looked like the parameters are in local space, which combined with our conventions document (cameras view along negative Z) led me to the wrong conclusion about which numbers go into `near` and `far`. This lead me down the rabbit hole that ultimately ended up in this discussion. 

What would have helped me avoid all this is some kind of note on how each value relates to the frustum's position in local space; hence me adding this note.

pylinalg/func/matrix.py
FirefoxMetzger: Sure. Shall we do this in this PR, or in a later one that touches `matrix_make_perspective`?

almarklein:
> What would have helped me avoid all this is some kind of note on how each value relates to the frustum's position in local space; hence me adding this note.

Ok, fair enough 😄

pylinalg/func/matrix.py
Korijn: You can do it whenever you like.

Korijn (approved):

PR #63, opened 2023-04-18
Keep dimensions when applying quaternions to vector batches

FirefoxMetzger:
Swatted another bug in pylinalg. When applying a quaternion to a batch of vectors via `vector_apply_quaternion` we need to keep the size-1 dimensions of the dot product internally to allow broadcasting to work correctly. Otherwise things can go sideways.

I added the fix as well as a regression test.


Korijn (dismissed):

FirefoxMetzger:
Now it is LGTM from my side as well (once the test suite is all green).

Korijn (approved):

PR #64, opened 2023-04-21
Fix quaternion from ax angle

FirefoxMetzger:
While working on the gizmo I realized that the axis passed to `quaternion_make_from_axis_angle` is not normalized when creating the rotation. As a result, the resulting quaternion isn't guaranteed to have unit length, meaning that we introduce a scaling alongside the rotation.

This PR:

1. Fixes this by normalizing the axis.
2. Upgrades the current unit test to test for various axis lengths.
3. Extends `quaternion_make_from_axis_angle` to allow broadcasting of inputs.
4. Adds `axis_angle_from_quaternion` to the `misc` module to convert quaternions back to axis-angle representation.
5. Adds a property-based test to ensure that we can convert from axis-angle to quaternion and back.

### DIFF
--- a/pylinalg/func/misc.py
+++ b/pylinalg/func/misc.py
@@ -1,7 +1,7 @@
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
 
-__all__ = ["aabb_to_sphere", "aabb_transform"]
+__all__ = ["aabb_to_sphere", "aabb_transform", "axis_angle_from_quaternion"]
 
 
 def aabb_to_sphere(aabb, /, *, out=None, dtype=None):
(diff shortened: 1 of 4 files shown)


Korijn (approved):

pylinalg/func/misc.py
```diff
@@ -107,3 +107,48 @@ def aabb_transform(aabb, matrix, /, *, out=None, dtype=None):
...
+    quaternion = np.asarray(quaternion)
+
+    if out is None:
+        quaternion = quaternion.astype(dtype)
+        out = (
+            quaternion[..., :3] / np.sqrt(1 - quaternion[..., 3] ** 2),
+            2 * np.arccos(quaternion[..., 3]),
+        )
+    else:
+        # Note: the elements of out must allow mutation, i.e., you can
```
almarklein, line 150: Looks like this sentence was not finished.

FirefoxMetzger: Thought I had deleted this sentence. It became a note in the docstring.

PR #65, opened 2023-04-21
Remove unneeded comment

FirefoxMetzger:


Korijn (approved):

Korijn:
I tried re-running CI a couple of times without success. It seems that #64 introduced some instability in the test suite:

```
=========================== short test summary info ============================
FAILED tests/func/test_quaternion.py::test_quaternion_from_axis_angle_roundtrip - hypothesis.errors.FailedHealthCheck:
It looks like your strategy is filtering out a lot of data. Health check found 50 filtered examples but only 9 good ones.
This will make your tests much slower, and also will probably distort the data generation quite a lot.
You should adapt your strategy to filter less. This can also be caused by a low max_leaves parameter in recursive() calls
See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this.
If you want to disable just this health check, add HealthCheck.filter_too_much to the suppress_health_check settings for this test.
================== 1 failed, 62 passed, 60 warnings in 11.06s ==================
```

FirefoxMetzger:
@Korijn Ah damn. I thought we could run a stricter unit test in #64 but it looks like hypothesis is too good at finding values that produce large numeric errors.

I will have a look.

FirefoxMetzger:
@Korijn CI is passing now. I separated two test scenarios to make it easier to generate valid example inputs (hypothesis was complaining that it was too hard to find good inputs), and now it seems that hypothesis is happy again.

Korijn (approved):

PR #66, opened 2023-04-25
Make quaternion from unit vectors broadcastable

FirefoxMetzger:
This PR extends the behavior of `quaternion_make_from_unit_vectors` to allow non-unit vectors as well as broadcasting. I tripped over this while porting pygfx (it says unit vectors in the name, but I thought it would simply normalize internally if needed), so I took the opportunity to generalize the behavior, make it broadcastable, and promote the unit test to a property-based one that will test more combinations of vectors.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -20,41 +20,46 @@ tests-no-zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy 
 
 [[package]]
 name = "black"
-version = "22.10.0"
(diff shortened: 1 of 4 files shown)


poetry.lock
Korijn, line 1: can you revert the changes to the lockfile?

FirefoxMetzger: there is a new version of black, which I updated to the latest version.

Korijn (approved):

PR #69, opened 2023-04-26
bump version and fix release step

Korijn:
* Bump version to 0.3.4
* Copy publish step from pygfx repo


.github/workflows/ci.yml
```diff
@@ -83,36 +83,30 @@ jobs:
...
         token: ${{ secrets.GITHUB_TOKEN }}
+        files: |
+          dist/*.tar.gz
+          dist/*.whl
     - name: Publish to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         user: __token__
-        password: ${{ secrets.PYPI_PASSWORD }}
+        password: ${{ secrets.PYPI_PASSWORD }}
```
FirefoxMetzger, line 112:
```suggestion
        password: ${{ secrets.PYPI_PASSWORD }}

```

Finish file with new line.

pylinalg/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 # flake8: noqa
 
-__version__ = "0.3.3"
+__version__ = "0.3.4"
```
FirefoxMetzger, line 9:
```suggestion
import pkg_resources

__version__ = pkg_resources.get_distribution('pylinalg').version
```

Shall we can reduce the places where we have to bump version numbers by 1?

FirefoxMetzger (approved):
Looks good 👍 Two small suggestions to make the repo a bit more tidy.

pylinalg/__init__.py
Korijn: Sure, but then we should use importlib as it is replacing pkg_resources. Let's do that in a future PR.

FirefoxMetzger (approved):

PR #72, opened 2023-05-02
Add `vector_euler_angles_from_quaternion`

### LINKED ISSUE #68: Add `vector_euler_angles_from_quaternion`
We can already go from Euler angles to quaternions, and it would also be good to be able to do the reverse/inverse. 

We could use this, for example, in pygfx to implement getters and setters for rotations on the transform classes, e.g. `obj.world.euler_rotation = (alpha, beta, gamma)` or `obj.world.euler_x = 42`.

Korijn:
Closes #68

I'm a little out of my depth here, I just took an implementation from stackoverflow and vectorized it.

I added two tests to compensate for my lack of confidence. 🤣

### DIFF
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -477,7 +477,7 @@ def vector_make_spherical_safe(vector, /, *, out=None, dtype=None):
 
 
 def vector_euler_angles_from_quaternion(quaternion, /, *, out=None, dtype=None):
-    """Convert quaternions to euler angles.
+    """Convert quaternions to XYZ euler angles.
 
     Parameters
     ----------
--- a/pylinalg/func/vector.py
+++ b/pylinalg/func/vector.py
@@ -476,4 +476,55 @@ def vector_make_spherical_safe(vector, /, *, out=None, dtype=None):
     return out
 
 
+def vector_euler_angles_from_quaternion(quaternion, /, *, out=None, dtype=None):
+    """Convert quaternions to euler angles.
+
+    Parameters
+    ----------
+    quaternion : ndarray, [4]
+        The quaternion to convert (in xyzw format).
+    out : ndarray, optional
+        A location into which the result is stored. If provided, it
+        must have a shape that the inputs broadcast to. If not provided or
+        None, a freshly-allocated array is returned. A tuple must have
+        length equal to the number of outputs.
+    dtype : data-type, optional
+        Overrides the data type of the result.
+
+    Returns
+    -------
+    out : ndarray, [3]
+        A set of XYZ euler angles.
+    """
+    quaternion = np.asarray(quaternion, dtype=float)
+
+    if out is None:
+        out = np.empty((*quaternion.shape[:-1], 3), dtype=dtype)
+
+    ysqr = quaternion[..., 1] ** 2
+
+    t0 = 2 * (
+        quaternion[..., 3] * quaternion[..., 0]
+        + quaternion[..., 1] * quaternion[..., 2]
+    )
+    t1 = 1 - 2 * (quaternion[..., 0] ** 2 + ysqr)
+    out[..., 0] = np.arctan2(t0, t1)
+
+    t2 = 2 * (
+        quaternion[..., 3] * quaternion[..., 1]
+        - quaternion[..., 2] * quaternion[..., 0]
+    )
+    t2 = np.clip(t2, a_min=-1, a_max=1)
+    out[..., 1] = np.arcsin(t2)
+
+    t3 = 2 * (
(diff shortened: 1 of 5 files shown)


pylinalg/func/vector.py
```diff
@@ -476,4 +476,55 @@ def vector_make_spherical_safe(vector, /, *, out=None, dtype=None):
     return out
 
 
+def vector_euler_angles_from_quaternion(quaternion, /, *, out=None, dtype=None):
```
almarklein, line 479: Why is this prefixed with `vector` if it returns scalar angles? If this fits the name-rules that we have adopted, then its fine, but it looks a bit odd for me on first sight.

almarklein (commented):
I cannot judge the implementation itself, so I looked mostly at the tests (and nitpicks).

One way the tests could be improved is to also test the batch-usage. This has been broken for other functions in the past. We can leave the hypothesis tests for the individual usage, and add a simple test similar to this:

https://github.com/pygfx/pylinalg/blob/a6688b79d9a374677444c2337d60ee5b47f32ffd/tests/func/test_vectors.py#L12-L37

tests/func/test_vectors.py
```diff
@@ -431,3 +431,46 @@ def test_vector_apply_matrix_orthographic():
...
+    order = "xyz"
+    quaternion = la.quaternion_make_from_euler_angles(angles, order=order)
+    matrix = la.matrix_make_rotation_from_euler_angles(angles, order=order)
+
+    angles_reconstructed = la.vector_euler_angles_from_quaternion(quaternion)
+    matrix_reconstructed = la.matrix_make_rotation_from_euler_angles(
+        angles_reconstructed
+    )
+
+    expected = la.vector_apply_matrix([1, 2, 3], matrix)
```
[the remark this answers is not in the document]
Korijn: Because they are not guaranteed to be equal. There are multiple valid reconstructions.

pylinalg/func/vector.py
```diff
@@ -476,4 +476,55 @@ def vector_make_spherical_safe(vector, /, *, out=None, dtype=None):
...
+    """Convert quaternions to XYZ euler angles.
+
+    Parameters
+    ----------
+    quaternion : ndarray, [4]
+        The quaternion to convert (in xyzw format).
+    out : ndarray, optional
+        A location into which the result is stored. If provided, it
+        must have a shape that the inputs broadcast to. If not provided or
```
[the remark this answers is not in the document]
almarklein: Oh, you're right! My bad.

pylinalg/func/vector.py
Korijn: Because it is a vector in the space of Euler space? I don't really have a good answer. I don't want to fuss about the names too much anymore as we'll rename everything to be less verbose soon anyway.

Korijn:
> I cannot judge the implementation itself, so I looked mostly at the tests (and nitpicks).
> 
> One way the tests could be improved is to also test the batch-usage. This has been broken for other functions in the past. We can leave the hypothesis tests for the individual usage, and add a simple test similar to this:
> 
> https://github.com/pygfx/pylinalg/blob/a6688b79d9a374677444c2337d60ee5b47f32ffd/tests/func/test_vectors.py#L12-L37

Sounds good, will do.

pylinalg/func/vector.py
almarklein: 👍

Korijn:
> > One way the tests could be improved is to also test the batch-usage. This has been broken for other functions in the past. We can leave the hypothesis tests for the individual usage, and add a simple test similar to [...]

Done!

tests/func/test_vectors.py
```diff
@@ -431,3 +431,69 @@ def test_vector_apply_matrix_orthographic():
...
+            la.quaternion_make_from_axis_angle([0, 0, 1], np.pi / 2),
+        ]
+    )
+    expected = np.array(
+        [
+            [np.pi, 0, 0],
+            [0, 0, 0],
+            [0, 0, np.pi / 2],
+        ]
+    )
```
almarklein, line 496: One of the concerns that I like this test to take away, is whether the shape of the output array is correct. I.e. whether walking over the sub-arrays one-by one produces the same result as batching them. However, the expected output is symmetric so exactly this kind of bug would slip through 😆. Adding one more case should fix that.

Korijn: I did as you asked and found another broadcasting support bug in `quaternion_make_from_axis_angle` which is now fixed as well (and covered by another unit test).

almarklein (approved):
Nice!

PR #73, opened 2023-05-05
Less verbose and cleaned up API

### LINKED ISSUE #67: Revise and clean public API
We want to revise and clean the public API as follows:

* Considering that the `pylinalg.obj` folder now only contains an empty init and that the "actual" object API lives in `pygfx.utils.transform`, we can remove the folder/submodule

* Drop `make_` in function names that include `from_`:

  To make function names a little less verbose, we discussed that it is possible to drop `make_` in functions that build objects `from_` some reference, e.g.

  - `quaternion_make_from_euler_angles` -> `quaternion_from_euler_angles`

  This loses no information while saving a view characters.

* We can make the API less verbose by introducing shorter function names and using namespaces to organize, e.g.:
  ```py
  cube.world.rotation = la.quat.mul(la.quat.from_euler((0.5, 0.01), order="XY"), cube.world.rotation)
  ```

Korijn:
Closes #67


pylinalg/__init__.py
```diff
@@ -12,5 +12,10 @@
...
 
-from .func import *
-from .obj import *
+from . import matrix, misc, quaternion, vector
+from .matrix import *
+from .misc import *
+from .quaternion import *
+from .vector import *
+
+__all__ = matrix.__all__ + misc.__all__ + quaternion.__all__ + vector.__all__
```
Korijn, line 21: I realize now this isn't as clean as it could be

pylinalg/vector.py
```diff
@@ -350,7 +350,7 @@ def vector_angle_between(vector_a, vector_b, /, *, out=None, dtype=None):
     return out
 
 
-def vector_from_matrix_position(homogeneous_matrix, /, *, out=None, dtype=None):
+def mat_decompose_translation(homogeneous_matrix, /, *, out=None, dtype=None):
```
almarklein, line 353: We could also choose`mat_decompose_position`. You probably have a better idea what makes the most sense. Commenting to make sure that this a conscious decision. Also see `mat_from_translation`.

pylinalg/__init__.py
almarklein: > I realize now this isn't as clean as it could be.

How do you mean? 

Is that definition of `__all__` here necessary, since we already make sure we clean `__all__`'s in the submodules? Or is it to help tooling?

almarklein (commented):
🚀

pylinalg/vector.py
Korijn: I think translation is better. A matrix may not be directly determining a position.

pylinalg/__init__.py
Korijn: I fixed it and added a test. If you approve of the changes, we can merge as far as I am concerned.

tests/test_pylinalg.py
```diff
@@ -0,0 +1,24 @@
...
+def test_api():
+    api = dir(la)
+
+    def popattr(key):
+        val = getattr(la, key)
+        api.remove(key)
+        return val
+
+    __version__ = popattr("__version__")
+    version_info = popattr("version_info")
```
almarklein, line 15:
```suggestion
    __version__ = popattr("__version__")  # noqa: F841
    version_info = popattr("version_info")  # noqa: F841
```

almarklein (approved):

PR #80, opened 2023-05-12
Add vertex_normals

Korijn:
I had a little bit of spare time and wanted to move this from `pygfx.utils.normals_from_vertices` to here to give it a proper place and documentation.

### DIFF
--- a/pylinalg/__init__.py
+++ b/pylinalg/__init__.py
@@ -5,6 +5,7 @@ Linear algebra utilities for Python.
 """
 from importlib.metadata import version
 
+from .geometry import *
 from .matrix import *
 from .misc import *
 from .quaternion import *
@@ -17,5 +18,7 @@ del version
 
 
 __all__ = [
-    name for name in globals() if name.startswith(("vec_", "mat_", "quat_", "aabb_"))
+    name
+    for name in globals()
+    if name.startswith(("vec_", "mat_", "quat_", "aabb_", "vertex_"))
 ]
--- a/pylinalg/geometry.py
+++ b/pylinalg/geometry.py
@@ -1,0 +1,68 @@
+import numpy as np
+
+
+def vertex_normals(vertices, indices, /, *, out=None, dtype=None):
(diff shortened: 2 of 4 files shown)


almarklein:
Is this linear algebra though? I am not against putting that here, but maybe it's good to discuss/decide what belongs in pylinalg and what not. E.g. do triangulation techniques also have a place here?

Also, while we're touching this code, would it make sense to support quads too? (Either by means of another function, or by detecting quads in the function itself.)

Korijn:
> Is this linear algebra though? I am not against putting that here, but maybe it's good to discuss/decide what belongs in pylinalg and what not. E.g. do triangulation techniques also have a place here?

The math involved is definitely linear algebra. I think it belongs. But I do not know where to "draw the line".


> Also, while we're touching this code, would it make sense to support quads too? (Either by means of another function, or by detecting quads in the function itself.)

I guess it does!

Korijn:
I've got heaps and heaps of functions that fall in this category that I would like to add here. Many of them are well suited/sufficiently generic for broad re-use. I'm a little tired of copy pasting them into all my projects. :)

I've had a think on this, and I feel like it definitely belongs, but perhaps we should create a subpackage to keep [computational geometry](https://en.wikipedia.org/wiki/Computational_geometry) functions grouped together, separately from the existing linear algebra functions.

What do you think of introducing a subpackage called `pylinalg.geom` or something, to keep it terse?

almarklein:
A subpackage could work. I think it all depends on how close we can consider such functions to linear algebra.

I think my main concern is that such functions look more like algorithms, and for some of these, Cython or Numba may be helpful to make the implementation performant, which might be a reason to bundle these in a separate package.

That said, if your collection of functions is all pure Python and you deem them close enough to pylinalg, I'll trust your judgement 😄

almarklein (approved):
> `pylinalg.geom`

I like `geometry` better.

almarklein:
> Any other ideas?

`pycgeo`, `pycompgeo` (with a p).

PR #82, opened 2023-07-04
Add `scale` keyword argument to `mat_decompose`

### LINKED ISSUE #81: Add `scale` keyword argument to `mat_decompose`
In order to preserve the scaling component in a roundtrip from components -> matrix -> components, we need to be able to pass a `scale` vector into `mat_decompose`

```py
def mat_decompose(matrix, /, *, scale=None, dtype=None, out=None):
```

Korijn:
Closes #81 

- [x] Add test exemplifying problematic roundtrip
- [x] Add `scale` keyword argument to `mat_decompose`

### DIFF
--- a/pylinalg/matrix.py
+++ b/pylinalg/matrix.py
@@ -364,9 +364,7 @@ def mat_decompose(matrix, /, *, scaling_signs=None, dtype=None, out=None):
     else:
         # if not, detect if a flip is needed to reconstruct the transform
         # and apply it to the first axis arbitrarily
-        flip = 1
-        if np.linalg.det(matrix) < 0:
-            flip = -1
+        flip = 1 if np.linalg.det(matrix) >= 0 else -1
         scaling_signs = np.array([flip, 1, 1])
 
     if out is not None:
--- a/pylinalg/matrix.py
+++ b/pylinalg/matrix.py
@@ -319,7 +319,7 @@ def mat_compose(translation, rotation, scaling, /, *, out=None, dtype=None):
     )
 
 
-def mat_decompose(matrix, /, *, dtype=None, out=None):
+def mat_decompose(matrix, /, *, scaling=None, dtype=None, out=None):
     """
     Decompose a transformation matrix into a translation vector, a
     quaternion and a scaling vector.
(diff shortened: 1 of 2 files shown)


almarklein:
Would it be sufficient to only provide the signs of the scaling?

Korijn:
> Would it be sufficient to only provide the signs of the scaling?

You are right. I've expanded the unit test to cover all possible enumerations of signs, and adjusted the function to only use the signs from the extra keyword argument.

pylinalg/matrix.py
```diff
@@ -353,13 +358,23 @@ def mat_decompose(matrix, /, *, dtype=None, out=None):
...
 
+    if scaling_signs is not None:
+        # if the user provides the scaling signs, always use them
+        scaling_signs = np.sign(scaling_signs)
+    else:
+        # if not, detect if a flip is needed to reconstruct the transform
+        # and apply it to the first axis arbitrarily
+        flip = 1
+        if np.linalg.det(matrix) < 0:
+            flip = -1
```
almarklein, line 369: Just a bit shorter.
```suggestion
        flip = 1 if np.linalg.det(matrix) >= 0 else -1
```

almarklein:
I wonder how we'd deal with cases where e.g. a pygfx both the object and parent have negative scaling signs. To create the world matrix, their matrices are multiplied. Should we separately multiply the `scale_signs`?

A related question, specific to the code here: should the function check whether the number of flipped axii matches the determinant? And what should it do if it does not match?

Korijn:
> I wonder how we'd deal with cases where e.g. a pygfx both the object and parent have negative scaling signs. To create the world matrix, their matrices are multiplied. Should we separately multiply the `scale_signs`?

I think we're only calling compose and decompose to construct the local matrix using the local SRT components, not on the world matrix. So that shouldn't be an issue.

> A related question, specific to the code here: should the function check whether the number of flipped axii matches the determinant? And what should it do if it does not match?

I thought about it but I don't think that's something you can glean off the value of the determinant, actually.

almarklein:
> I thought about it but I don't think that's something you can glean off the value of the determinant, actually.

Would it not be something like, if the det is smaller than zero, that the number of negative axii should be uneven?

Korijn:
I've added the validation as far as I think we can provide it, and an accompanying unit test.

almarklein (approved):
Cool, let's see how this works out in pygfx!

PR #84, opened 2023-10-19
Rename 2 funcs eucledian -> eucledean

almarklein:


Korijn (approved):

almarklein:
BTW, we're not using these functions in pygfx, we only refer to them in a comment.

Issue #86, opened 2024-05-21
Conversion between euler and quats doesn't seem stable


rvanlaar:
While creating a pygfx Controller I noticed the following:

Moving from euler to quats and back is not stable.
Here's an example:
```
In [11]: la.quat_to_euler(la.quat_from_euler(np.array([math.pi/2, math.pi/180,0])))
Out[11]: array([1.57079633e+00, 1.73472348e-18, 1.74532925e-02])

Almost as if the last two items of the array are reversed.

```

In code:
```
import math
import numpy as np
import pylinalg as la

la.quat_to_euler(la.quat_from_euler(np.array([math.pi/2,math.pi/180,0])))
```

rvanlaar:
Looked at scipy, and that one works:

```
import numpy as np
import pylinalg as la

from scipy.spatial.transform import Rotation


quat = np.array([0.70707986, 0.00617059, 0.00617059, 0.70707986])

la.quat_to_euler(quat)
Rotation.from_quat(quat).as_euler("XYZ")
```

Korijn:
Would you mind opening a pull request with a unit test that demonstrates the failure?

rvanlaar:
Done: https://github.com/pygfx/pylinalg/pull/87

It seems the issue has to do with the order. The default uppercase "XYZ" has a different result than lowercase "xyz".

Korijn:
https://github.com/pygfx/pylinalg/blob/92a83a223d4580c4438d54c78d23387f781f3faa/pylinalg/quaternion.py#L296

Using uppercase letters apparently makes that axis extrensic. I don't really understand what that means.

rvanlaar:
```
The three rotations can either be in a global frame of reference (extrinsic) or in a 
body centred frame of reference (intrinsic), which is attached to, and moves with, the object under rotation 
```

https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.from_euler.html

My 2cts: 
- Document the behavior
- give quat_to_euler also an order
- have quat_to_euler and quat_from_euler use the same default order.

Korijn:
If you want to take a shot at it, you're more than welcome. We're shorthanded at the moment.

rvanlaar:
Thanks for the offer. Am very busy, but who knows.

PR #87, opened 2024-05-22
Test for UPPERcase bug #86

rvanlaar:
quats_from_euler are handled differently when using uppercase vs lowercase order. The default is uppercase which doesn't give the right angles when converting back to euler.


Korijn:
See #94

rvanlaar:
I missed this in December. Thanks for resolving, appreciate it.

PR #88, opened 2024-07-07
Avoid deprecation warning in pytest

hmaarrfk:
It seems that they don't like this anymore. It doesn't seem used.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ from hypothesis.extra.numpy import arrays, from_dtype
 import pylinalg as la
 
 
-def pytest_report_header(config, start_path, startdir):
+def pytest_report_header(config):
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Just FYI:
```
tests/conftest.py:13
  /home/mark/git/pylinalg/tests/conftest.py:13: PytestRemovedIn9Warning: The (startdir: py.path.local) argument is deprecated, please use (start_path: pathlib.Path)
  see https://docs.pytest.org/en/latest/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path
    def pytest_report_header(config, start_path, startdir):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

Korijn (approved):

PR #90, opened 2024-12-19
fix decompose when scaling is 0

panxinmiao:
When performing matrix decomposition, if any scaling factor is zero, a division by zero exception occurs, leading to NaN values in the matrix. This causes subsequent SVD computations to fail.

When the scaling factor along a certain axis is zero, the rotation along that axis becomes meaningless. In this case, the rotational component along that axis should be ignored, and the corresponding column in the rotation matrix (representing that axis' rotation) should be set to zero.

See: [Interpolation Test](https://github.com/KhronosGroup/glTF-Sample-Assets/blob/main/Models/InterpolationTest/README.md)  case in https://github.com/pygfx/pygfx/issues/910

### DIFF
--- a/pylinalg/matrix.py
+++ b/pylinalg/matrix.py
@@ -379,7 +379,11 @@ def mat_decompose(matrix, /, *, scaling_signs=None, dtype=None, out=None):
     scaling *= scaling_signs
 
     rotation = out[1] if out is not None else None
-    rotation_matrix = matrix[:-1, :-1] * (1 / scaling)[None, :]
+
+    rotation_matrix = matrix[:-1, :-1].copy()
+    mask = scaling != 0
+    rotation_matrix[:, mask] /= scaling[mask][None, :]
+    rotation_matrix[:, ~mask] = 0.0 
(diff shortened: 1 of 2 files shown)


Korijn:
Can you add a unit test?

Korijn (approved):

Korijn:
Some linting failures, can you address those?

```sh
./pylinalg/matrix.py:386:36: BLK100 Black would make changes.
./pylinalg/matrix.py:386:36: W291 trailing whitespace
./tests/test_matrix.py:210:1: BLK100 Black would make changes.
./tests/test_matrix.py:210:1: E302 expected 2 blank lines, found 1
```

And to solve the other failure we'll have to patch the CI pipeline unfortunately...

```sh
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/
```

panxinmiao:
> Some linting failures, can you address those?
> 

Updated :)

PR #94, opened 2024-12-27
Fix order argument for quat_to_euler, quat_from_euler & mat_from_eulet

### LINKED ISSUE #86: Conversion between euler and quats doesn't seem stable

Korijn:
Closes #86

This was way complicated. Thank goodness there was a robust python implementation available.

Changes:

* Adds order argument to `quat_to_euler`
* Correct support for intrinsic and extrinsic rotations
* ⚠️ **breaking change** Aligns order argument convention with scipy's: extrinsic is now lowercase, it was uppercase before.
* Bump version to 0.6.0 because of the breaking changes
* Improved order argument docstring.
* Lots of new tests, including the case from the bug report, and comparisons to scipy's implementation

### DIFF
--- a/pylinalg/matrix.py
+++ b/pylinalg/matrix.py
@@ -1,5 +1,3 @@
-from math import cos, sin
-
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
 
@@ -129,8 +127,9 @@ def mat_from_euler(angles, /, *, order="xyz", out=None, dtype=None):
     angles : ndarray, [3]
         The euler angles.
     order : string, optional
-        The order in which the rotations should be applied. Default
-        is "xyz".
+        The rotation order as a string. Can include 'X', 'Y', 'Z' for intrinsic
+        rotation (uppercase) or 'x', 'y', 'z' for extrinsic rotation (lowercase).
+        Default is "xyz".
     out : ndarray, optional
         A location into which the result is stored. If provided, it
         must have a shape that the inputs broadcast to. If not provided or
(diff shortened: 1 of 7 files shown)


pylinalg/matrix.py
```diff
@@ -155,31 +154,42 @@ def mat_from_euler(angles, /, *, order="xyz", out=None, dtype=None):
...
+            affine_matrix[1, 0] = np.sin(angle)
+            affine_matrix[1, 1] = np.cos(angle)
+
+        if fill_out_first:
+            out[:] = affine_matrix
+            fill_out_first = False
+        elif out is None:
+            out = affine_matrix
+        else:
+            out @= affine_matrix
```
Korijn, line 191: Probably just echoing into the void here but this was the first time in my life I used this operator.

pylinalg/matrix.py
```diff
@@ -129,8 +127,9 @@ def mat_from_euler(angles, /, *, order="xyz", out=None, dtype=None):
     angles : ndarray, [3]
         The euler angles.
     order : string, optional
-        The order in which the rotations should be applied. Default
-        is "xyz".
+        The rotation order as a string. Can include 'X', 'Y', 'Z' for intrinsic
+        rotation (uppercase) or 'x', 'y', 'z' for extrinsic rotation (lowercase).
+        Default is "xyz".
```
almarklein, line 132: Out of curiosity, can a user also pass (or think he can pass) a mix of lowercase and uppercase?

Korijn: Yeah they can but it will not give well defined results. I could validate inputs better.

PR #97, opened 2025-01-07
Add mat_inverse method

panxinmiao:
Following @Korijn 's suggestion in https://github.com/pygfx/pygfx/pull/919#issuecomment-2566318478, added the `mat_inverse` method to pylinalg.


pylinalg/matrix.py
```diff
@@ -631,6 +631,242 @@ def mat_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
...
+        the number of outputs.
+
+    Returns
+    -------
+    out : ndarray, [4, 4]
+        The inverse of the input matrix.
+
+    Notes
+    -----
+    The default method is "numpy" when numpy version > 2.0.0,
```
Korijn, line 842: >=

Korijn:
Nice work! Just need to address the merge conflicts and the one review comment, then we can merge. :)

pylinalg/matrix.py
```diff
@@ -631,6 +631,242 @@ def mat_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
     return out
 
 
+def _mat_inv(m):
+    # Reference:
+    # https://github.com/mrdoob/three.js/blob/dev/src/math/Matrix4.js
+    # based on:
+    # http://www.euclideanspace.com/maths/algebra/matrix/functions/inverse/fourD/index.htm
+    out = np.empty((4, 4), dtype=m.dtype)
```
Korijn, line 639: I think the return type of np.linalg.inv is always float, so it's fine to do that here as well.

pylinalg/matrix.py
```diff
@@ -673,6 +673,242 @@ def mat_look_at(eye, target, up_reference, /, *, out=None, dtype=None):
...
+    }[method]
+
+    matrix = np.asarray(matrix)
+    if out is None:
+        out = np.empty_like(matrix, dtype=dtype)
+    try:
+        out[:] = fn(matrix)
+    except np.linalg.LinAlgError:
+        out[:] = 0
+    return out
```
Korijn, line 909: I think with this strategy we can avoid one matrix allocation in the most common use cases:

```suggestion
    matrix = np.asarray(matrix)
    try:
        inverse = fn(matrix)
    except np.linalg.LinAlgError:
        inverse = np.zeros_like(matrix, dtype=dtype)
    if out is None:
        if dtype is not None:
            return inverse.astype(dtype, copy=False)
        return inverse
    else:
        out[:] = inverse
    return out
```

Korijn (approved):

almarklein:
Nice work! A comment and a question:

* The method option "manual" strikes me as a bit odd. It's not really manual, but pure Python.
* Would it be good to use this new inv method? If I checked correctly, we (only) use it in `vec_unproject()`

Korijn:
> * The method option "manual" strikes me as a bit odd. It's not really manual, but pure Python.
> * Would it be good to use this new inv method internally? If I checked correctly, we (only) use it in `vec_unproject()`

Didn't think about that! I will release a patch.

Korijn:
Done in #100

PR #100, opened 2025-01-10
Various improvements

Korijn:
* Rename matrix inversion method "manual" to "python"
* Reduce a few more matrix allocations
* Add option to mat_inverse to raise an error if the matrix is invertible. The default remains returning all zeros.
* Add option to vec_unproject to skip matrix inversion.
* Bump version


almarklein (approved):
Nice!

PR #104, opened 2025-02-23
🔥 Optimize vec_transform again

Korijn:
Follow-up to #103

* The code felt messy so I wanted to clean it up and make it DRY
* I felt like the `projection=True` code could be faster too, and it did measurably become a bit faster
* Turns out leaving out the up-front type conversions leads to the same precision and performs as well or better

Average per call timings:

```
# N=100000
vec_transform (projection=True) 3.6813059999258257e-06
vec_transform (projection=False) 2.56878800020786e-06
apply_matrix 2.81830899999477e-06

# N=1000
vec_transform (batch, projection=True) 0.002807431799999904
vec_transform (batch, projection=False) 0.002085096800001338
apply_matrix (batch) 0.0020407788000011353
```

### DIFF
--- a/pylinalg/vector.py
+++ b/pylinalg/vector.py
@@ -105,28 +105,51 @@ def vec_transform(
         transformed vectors
     """
 
-    matrix = np.asarray(matrix, dtype="f8")
+    matrix = np.asarray(matrix)
+    vectors = np.asarray(vectors)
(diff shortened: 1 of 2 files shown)


Korijn:
This been released as `v0.6.6`

PR #106, opened 2025-02-24
Run the pygfx test suite as part of CI

Korijn:


Korijn:
@almarklein perhaps this is also useful for rendercanvas... and maybe even wgpu-py?

almarklein:
What an interesting idea 🤔

